PR #28, opened 2026-04-08
feat: add BitcoinSX (.sx) output backend

icellan:
## Summary

Adds a parallel compilation backend that emits [BitcoinSX](https://www.bitcoinsx.io/docs) (.sx) text from the ANF IR, preserving high-level structure that the hex backend destroys during stack lowering.

- Private methods become SX `#macros` with bare-name invocation
- Bounded loops become SX `repeat Nn ... end` blocks (not unrolled)
- If/else becomes SX `if ... else ... endIf`
- Constructor parameters become SX `.variable` placeholders
- All opcodes use SX camelCase names (`dup`, `checkSig`, `hash160`)
- Stateful contract intrinsics (checkPreimage, deserializeState, addOutput, extractors) are fully supported
- Specialized codegen (EC, SHA-256, SLH-DSA, BabyBear, Merkle) bridges through StackOp→SX conversion

### Example output

P2PKH:
```sx
// P2PKH.sx — generated by Runar compiler
//
// Constructor:
//   .pubKeyHash: Addr

// Public method: unlock(sig: Sig, pubKey: PubKey)
dup
hash160
.pubKeyHash
equal
verify
checkSig
```

Contract with private method:
```sx
#requireOwner
  // params: sig: Sig
  .owner
  checkSig
  verify
end

// Public method: spend(sig: Sig)
requireOwner
true
```

Contract with loop:
```sx
0n
repeat 3n
  2n pick
  rot
  add
  over
  add
  swap
  1add
end
drop
.expected
numEqual
```

## Architecture

The SX backend is a new pass 7 that operates on the same ANF IR as the existing hex backend (passes 5+6):

```
                                    ┌─ 05-stack-lower → 06-emit → hex/asm (existing)
ANF IR (Pass 4 output) ──┤
                                    └─ 07-sx-emit (NEW) → .sx text
```

**New files:**
- `packages/runar-compiler/src/passes/07-sx-emit.ts` — Main emitter: orchestrates macros, dispatch, method bodies
- `packages/runar-compiler/src/passes/07-sx-stack.ts` — SX stack scheduler: tracks named values on virtual stack, emits SX tokens, handles all 17 ANF value kinds
- `packages/runar-compiler/src/passes/07-sx-codegen-bridge.ts` — OP→SX camelCase name table (80+ opcodes) + StackOp-to-SX text converter
- `packages/runar-compiler/src/ir/sx-ir.ts` — SX output types
- `packages/runar-compiler/src/__tests__/07-sx-emit.test.ts` — 41 tests

**Modified files:**
- `packages/runar-compiler/src/passes/05-stack-lower.ts` — Exported `lowerBindingsToOps()` helper + `stackDepth`/`peekStack` accessors, allowing the SX backend to delegate complex intrinsics to the proven hex backend without code duplication
- `packages/runar-compiler/src/index.ts` — Added `emitSX` compile option, wired pass 7 into pipeline
- `packages/runar-compiler/src/ir/artifact.ts` + `assembler.ts` — Added `sx?: string` to `RunarArtifact`
- `packages/runar-compiler/src/ir/index.ts` — Exported SX types
- `packages/runar-cli/src/bin.ts` + `commands/compile.ts` — Added `--sx` CLI flag

### Key design decisions

**Delegation over duplication:** Stateful contract intrinsics (checkPreimage, deserializeState, addOutput, output hash verification, preimage field extractors) and complex math builtins (pow, sqrt, gcd, log2, etc.) delegate to the existing hex backend via `lowerBindingsToOps()` → `stackOpsToSX()` bridge. This avoids duplicating ~1000 lines of carefully tested opcode sequences while producing correct SX output for all contract types.

**SX peephole optimizer:** Removes redundant token patterns (`swap swap`, `dup drop`, `false drop`) from the SX output.

**Macro calling convention:** Before calling a `#macro`, all arguments are pushed onto the stack in parameter order. The macro body assumes args at known positions relative to stack top, and leaves its result (if any) on top.

**Loop counter strategy:** SX `repeat` doesn't provide an implicit loop counter. When the loop body references `iterVar`, the scheduler pushes a counter before the repeat block, increments at the end of each iteration body, and drops after `end`.

**Non-fatal emission:** SX emission failure produces a warning diagnostic, not an error. The hex output is always produced regardless of SX success.

## CLI usage

```bash
runar compile MyContract.runar.ts --sx              # writes .sx + artifact JSON
runar compile MyContract.runar.ts --sx -o build/    # custom output dir
runar compile MyContract.runar.ts --sx --asm        # SX file + ASM to stdout
```

## Test plan

- [x] 41 new tests in `07-sx-emit.test.ts` covering all contract patterns
- [x] Full test suite: 1493/1493 passing (including 136 cross-compiler conformance)
- [x] Type-checks clean with `tsc --noEmit`
- [ ] Verify SX output parses in the [BitcoinSX IDE](https://www.bitcoinsx.io)
- [ ] Test with larger contracts from `examples/ts/`


icellan:
Closing as stale. This adds a new TS-only BitcoinSX (.sx) output backend (+2364/-3) but has sat since April and is now 700+ commits behind `main` with unknown mergeability. Whether Rúnar ships a .sx backend is a product decision rather than a fix, so it shouldn't be merged opportunistically. The branch (`feature/sx-backend`) is preserved on the remote — reopen and rebase onto current `main` if the BitcoinSX backend is something the project wants to pursue.

PR #35, opened 2026-04-28
fix(compiler): issue #34 — cross-method param shadowing breaks byte-type analysis (+ pow runtime guard)

### LINKED ISSUE #34: Compiled bytecode diverges from interpretive: hand-unrolled `if` iterations + nested `let` reassignment
# Possible compiler/codegen divergence: hand-unrolled `if` iterations

**Versions**

- `runar-compiler@0.4.4`
- `runar-lang@0.4.4`
- `runar-testing@0.4.4`
- `@bsv/sdk@2.0.13`
- `node v22.22.2`

## Summary

A two-method contract whose body uses 8 hand-unrolled `if (Nn < count) { ... }`
iterations — each combining a `let` reassignment of a method-scope
pointer with a nested `let` reassignment via `if (blob === target) { found = true }` —
compiles to bytecode that:

- **passes** under `runar-testing`'s `TestContract.fromSource` (interpretive).
- **fails** under `@bsv/sdk`'s `Spend.validate` (compiled bytecode through the
  Bitcoin Script interpreter) with `OP_SPLIT` on the second iteration when
  the iter-1 bound check `if (1n < count)` should be skipped (`count = 1`).

Decoded compiled bytecode shows the iter-1 bound check picks
`count` from the same depth as iter 0 (`OP_4 OP_PICK`), but iter 0's
body has changed the stack shape (substr ops, the conditional `let found = true`
write, and the `let p = p + 1 + x` reassignment), so the depth-4
PICK at iter 1 reads a stale item rather than `count`. The bound
check then sees `1 < <stale>`, takes the branch, and the iter-1
substr fires `OP_SPLIT` on a corrupted offset.

I'm not certain whether this is a compiler bug or an unsupported
combination of features I should have known to avoid — happy to
adapt either way.

### Source — `StackTrackerRepro.runar.ts`

```ts
import { SmartContract, assert, substr, bin2num, num2bin, cat } from 'runar-lang';
import type { ByteString } from 'runar-lang';

class StackTrackerRepro extends SmartContract {
  constructor() { super(); }

  public walk(buf: ByteString, count: bigint, target: ByteString) {
    let p: bigint = 0n;
    let found: boolean = false;

    if (0n < count) {
      const x: bigint = bin2num(cat(substr(buf, p, 1n), num2bin(0n, 1n)));
      const blob: ByteString = substr(buf, p, 1n + x);
      if (blob === target) { found = true; }
      p = p + 1n + x;
    }
    if (1n < count) {
      const x: bigint = bin2num(cat(substr(buf, p, 1n), num2bin(0n, 1n)));
      const blob: ByteString = substr(buf, p, 1n + x);
      if (blob === target) { found = true; }
      p = p + 1n + x;
    }
    if (2n < count) {
      const x: bigint = bin2num(cat(substr(buf, p, 1n), num2bin(0n, 1n)));
      const blob: ByteString = substr(buf, p, 1n + x);
      if (blob === target) { found = true; }
      p = p + 1n + x;
    }
    if (3n < count) {
      const x: bigint = bin2num(cat(substr(buf, p, 1n), num2bin(0n, 1n)));
      const blob: ByteString = substr(buf, p, 1n + x);
      if (blob === target) { found = true; }
      p = p + 1n + x;
    }
    if (4n < count) {
      const x: bigint = bin2num(cat(substr(buf, p, 1n), num2bin(0n, 1n)));
      const blob: ByteString = substr(buf, p, 1n + x);
      if (blob === target) { found = true; }
      p = p + 1n + x;
    }
    if (5n < count) {
      const x: bigint = bin2num(cat(substr(buf, p, 1n), num2bin(0n, 1n)));
      const blob: ByteString = substr(buf, p, 1n + x);
      if (blob === target) { found = true; }
      p = p + 1n + x;
    }
    if (6n < count) {
      const x: bigint = bin2num(cat(substr(buf, p, 1n), num2bin(0n, 1n)));
      const blob: ByteString = substr(buf, p, 1n + x);
      if (blob === target) { found = true; }
      p = p + 1n + x;
    }
    if (7n < count) {
      const x: bigint = bin2num(cat(substr(buf, p, 1n), num2bin(0n, 1n)));
      const blob: ByteString = substr(buf, p, 1n + x);
      if (blob === target) { found = true; }
      p = p + 1n + x;
    }

    assert(found);
  }

  public other(x: ByteString) { assert(x === x); }
}

export default StackTrackerRepro;
```

### Inputs

```
buf    = 05aaaaaaaaaabbbbbbbbbbbbbbbbbbbb   (16 bytes)
count  = 1n
target = 05aaaaaaaaaa                       (the 6-byte record at buf[0..6])
```

For `count = 1n`, only iter 0 should execute; iters 1–7 should be
skipped by their bound checks. Iter 0 reads `x = 0x05`, builds
`blob = buf[0..6]`, matches `target`, sets `found = true`,
advances `p = 6`. The trailing `assert(found)` should pass.

### Interpretive (passes)

```ts
const c = TestContract.fromSource(SOURCE, {});
const r = c.call('walk', { buf, count: 1n, target });
// r.success === true
```

### Compiled bytecode (fails)

```ts
const sec = ScriptExecutionContract.fromSource(SOURCE, {});
const lockingScript = LockingScript.fromHex(sec.scriptHex);

const unlocking = new BsvScript();
unlocking.writeBin(Array.from(hexToBytes(buf)));     // arg 0: buf
unlocking.writeOpCode(0x51);                          // arg 1: count = 1n (OP_1)
unlocking.writeBin(Array.from(hexToBytes(target)));   // arg 2: target
unlocking.writeOpCode(0x00);                          // method selector = walk (OP_0)

const spend = new BsvSpend({
  sourceTXID: 'cd'.repeat(32),
  sourceOutputIndex: 0,
  sourceSatoshis: 1000,
  lockingScript,
  transactionVersion: 1,
  otherInputs: [],
  outputs: [/* any P2PKH placeholder */],
  inputIndex: 0,
  unlockingScript: UnlockingScript.fromHex(unlocking.toHex()),
  inputSequence: 0xffffffff,
  lockTime: 0,
  isRelaxed: true,
});
spend.validate();
```

throws:

```
ScriptEvaluationError: Script evaluation error:
OP_SPLIT requires the first stack item to be a non-negative number
less than or equal to the size of the second-from-top stack item.
TXID:    cdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcd
Context: LockingScript, PC: 40
Stack:   [05aaaaaaaaaabbbbbbbbbbbbbbbbbbbb, 01, 05aaaaaaaaaa, , , 05] (len: 6, mem: 24)
AltStack: []
IfStack:  [true, true]
```

The 6-item stack at error and `IfStack: [true, true]` indicate
execution entered the iter-1 body even though `count = 1n`.

## Decoded bytecode (relevant prefix)

```
OP_DUP OP_0 OP_NUMEQUAL OP_IF OP_DROP                 ← method dispatch
OP_0 OP_FALSE                                         ← let p = 0n; let found = false
OP_0 OP_4 OP_PICK OP_LESSTHAN OP_IF                   ← if (0n < count)  — count picked at depth 4
  OP_4 OP_PICK OP_1 OP_SWAP OP_3 OP_PICK
  OP_SPLIT OP_NIP OP_SWAP OP_SPLIT OP_DROP            ← substr(buf, p, 1)
  OP_0 OP_1 OP_NUM2BIN OP_CAT OP_BIN2NUM              ← bin2num(... || 0x00) → x
  OP_5 OP_PICK OP_1 OP_2 OP_PICK OP_CAT
  OP_SWAP OP_4 OP_PICK
  OP_SPLIT OP_NIP OP_SWAP OP_SPLIT OP_DROP            ← substr(buf, p, 1+x) → blob
  OP_4 OP_PICK OP_EQUAL
  OP_IF OP_TRUE OP_ELSE OP_OVER OP_ENDIF
  OP_2 OP_ROLL OP_DROP                                ← found = (blob === target) ? true : found
  OP_1 OP_3 OP_PICK OP_SWAP OP_ADD OP_ROT OP_CAT      ← p = p + 1 + x
OP_ELSE OP_OVER OP_ENDIF OP_2 OP_ROLL OP_DROP

OP_1 OP_4 OP_PICK OP_LESSTHAN OP_IF                   ← if (1n < count) — STILL OP_4 OP_PICK
  ...
```

Iter 0's body modifies the stack (substr ops, conditional `found` write,
`p` reassignment) but iter 1's bound check still uses `OP_4 OP_PICK`,
so the picked value at iter 1 isn't `count` anymore.

## Observations

- Single-method contracts don't appear to trigger this — both methods
  are needed (presumably because the method-dispatch preamble shifts
  what's on the stack at iter 0).
- Reducing to 4 hand-unrolled iters made the test pass; 8 reproduces.
  Haven't fully bisected the minimum.
- Removing the nested `if (blob === target) { found = true }` write
  (leaving only `let p` reassignment) also makes the test pass.
- `MAX_POW_ITERATIONS = 32` in
  `passes/05-stack-lower.js` (line 2912) silently caps `pow(2n, exp)` at
  `exp = 32`, separately from the issue above. Worth a consumer-side
  warning regardless of whether this report turns out to be a real bug.

## Question

Is the depth tracking across hand-unrolled `if` blocks with
inside-block `let` reassignment expected to work, or is there a
recommended pattern I should follow instead? Happy to either work
around this on my side or land a fix if you'd like a PR.

icellan:
## Summary

Fixes [#34](https://github.com/icellan/runar/issues/34) across all 7 compilers, plus the secondary `MAX_POW_ITERATIONS = 32` silent-cap bug surfaced in the same report.

### Root cause

`getParamType` (and its peers in the 6 mirror compilers) searched the constructor + ALL methods for a parameter name match. When two methods shared a parameter name — e.g. the issue's repro had `walk(buf, count, target)` plus `other(x: ByteString)` — references to the local `x` (a `bigint` result of `bin2num`) inside `walk` falsely matched `other`'s `x: ByteString` parameter, marking the local as byte-typed.

Downstream effect: `lowerBinaryExpr` set `result_type: 'bytes'` on `1n + x` (since one operand was "byte-typed"), which made the stack-lowering pass emit `OP_CAT` for the addition. The corrupted length operand passed to a subsequent `OP_SPLIT` then failed with the original report's symptom:

```
OP_SPLIT requires the first stack item to be a non-negative number
less than or equal to the size of the second-from-top stack item.
```

The user's hypothesis (depth tracking across hand-unrolled `if` blocks) was a downstream symptom: the corrupted byte-string operand only arrived at the failing `OP_SPLIT` after the inner-`if` + `let`-reassign body executed, which made the failure look like a stack-depth desync.

### Fix

- Introduce a method-scoped `methodParamTypes` table on every compiler's lowering context, populated by `setMethodParamTypes(params)` at the start of each method (and the constructor).
- `getParamType(name)` now consults this table only — no more cross-method lookup.

Mirrored byte-identical across:
- `packages/runar-compiler/src/passes/04-anf-lower.ts`
- `compilers/go/frontend/anf_lower.go`
- `compilers/rust/src/frontend/anf_lower.rs`
- `compilers/python/runar_compiler/frontend/anf_lower.py`
- `compilers/zig/src/passes/anf_lower.zig`
- `compilers/ruby/lib/runar_compiler/frontend/anf_lower.rb`
- `compilers/java/src/main/java/runar/compiler/passes/AnfLower.java`

### Bonus: `pow` runtime guard

The same report flagged that `MAX_POW_ITERATIONS = 32` silently saturates `pow(base, exp)` at `base^32` for non-literal `exp`. Bitcoin Script can't loop, so the cap is structurally required — but silent saturation is a real correctness bug. Added a runtime guard at the start of the unrolled pow loop:

```
<exp> <MAX> OP_LESSTHANOREQUAL OP_VERIFY
```

Script aborts cleanly on `exp > 32` instead of returning the wrong number. Mirrored across all 7 compilers; goldens regenerated for `math-demo`.

### Conformance fixture

Added `conformance/tests/nested-if-multi-reassign/` (the issue's exact repro contract) so all 7 compilers golden-test it going forward.

### Companion fixes

- `01-parse-java.test.ts`: drop unused `MemberExpr` import (TS6133) that blocked `pnpm --filter runar-compiler build` in a fresh worktree.
- `zig-parser-examples.test.ts`: register the new fixture (and the long-pending `multisig-2of3`) under `ZIG_PORT_PENDING` until each Zig port lands.
- `script-execution.test.ts`: fix the `readContract` helper's stale path (it expected `<name>.runar.ts` directly inside the conformance dir, but contracts live under `examples/<lang>/<name>/<Name>.runar.ts`); now reads via the fixture's `source.json`.

## Test plan

- [x] `npx vitest run script-execution -t "issue #34"` — RED before, GREEN after.
- [x] `npx vitest run` — no new TS failures (PriceBet + MultiSig2of3 cross-compiler are pre-existing on `main`).
- [x] `cd compilers/go && go test ./...` — all 5 packages pass.
- [x] `cd compilers/rust && cargo test` — 22 + 64 + … pass.
- [x] `cd compilers/python && python3 -m pytest` — 731 pass.
- [x] `cd compilers/zig && zig build test` — 523/523 pass.
- [x] `cd compilers/ruby && rake test` — 20 runs, 44 assertions, 0 failures.
- [x] `cd compilers/java && gradle clean test` — BUILD SUCCESSFUL.

Closes #34


icellan:
Superseded by a fresh re-implementation against current `main`: this PR is 440 commits behind and conflicts in the core fix (7-tier `anf_lower` + the `math-demo` golden). The underlying bug is still live on `main` (`getParamType` still searches all methods), so it has been re-implemented and fully verified (all 7 tiers byte-identical, conformance 62/62) in a new PR. Thanks for the original diagnosis + repro — the new fixture is based on it.

PR #43, opened 2026-05-16
runar-rs: fix BIP-143 sighash subscript trim for terminal methods

### LINKED ISSUE #42: Terminal methods on `StatefulSmartContract` produce invalid BIP-143 sighash subscript
## Summary

The Rust SDK's BIP-143 sighash-subscript construction for **terminal methods** on `StatefulSmartContract` subclasses computes a subscript that does not match what the on-chain script will hash at validation time.

The on-chain script trims its subscript at the method's `OP_CODESEPARATOR` position (per BIP-143). The SDK builds the sighash using the wrong subscript range for terminal methods (e.g. `Auction.close`), producing a signature that fails CHECKSIG with `arc error 461 — Signature must be zero for failed CHECK(MULTI)SIG operation` (NULLFAIL) on mainnet ARC broadcasters.

The state-mutating `bid()` method (different code path) works correctly — only the terminal `close()` path is affected.

## Why this isn't surfaced by existing tests

`integration/rust/tests/auction.rs::test_auction_close`:
1. Uses `deadline = 0` (see #40 — `nLockTime` is hardcoded to 0, so close requires deadline=0 to even attempt).
2. Runs against regtest via `RPCProvider::new_regtest()`, which has looser script-validation policy than mainnet ARC broadcasters.

When the locktime fix from #40 enables a non-zero deadline, and the test is broadcast against a mainnet ARC endpoint (TAAL, GorillaPool, etc.), the close call returns NULLFAIL.

## Affected files

- `packages/runar-rs/src/sdk/contract.rs` (subscript-trim conditional + `get_code_sep_index` helper)

The TS SDK has the same gap in `packages/runar-sdk/src/contract.ts`.

## Root cause

Two interlocking conditions:

1. **Subscript trim condition over-narrowly gated on `is_stateful`**: the BSV consensus rule (BIP-143 derived) is that any `OP_CODESEPARATOR` in the script before `OP_CHECKSIG` resets the subscript regardless of whether the contract has state fields. The SDK was only applying the trim when `is_stateful` was true.

2. **`get_code_sep_index` falls back to a legacy adjusted offset** when invoked through `from_txid` with the actual on-chain script (rather than a synthetic test deploy). For terminal methods, the index returned wasn't the actual byte position of the method's `OP_CODESEPARATOR`.

## Proposed fix

Two-part fix to `packages/runar-rs/src/sdk/contract.rs`:

**Part 1** — drop the `is_stateful` term from the trim condition. Trim whenever there's an `OP_CODESEPARATOR` in the script (`code_sep_idx >= 0`), regardless of whether the contract is stateful. Per BIP-143 + BSV consensus, this is the correct condition.

**Part 2** — replace `get_code_sep_index`'s legacy fallback with a robust `find_codesep_offsets` walker that:

- Walks the actual on-chain locking script byte-by-byte
- Correctly handles all BSV push opcodes (0x01..0x4b, OP_PUSHDATA1/2/4)
- Returns the actual byte position of each `OP_CODESEPARATOR`

## Evidence

A patched SDK produces signatures that validate against on-chain CHECKSIG for terminal methods. A test transaction broadcast to BSV mainnet mines successfully ([`779ea679…3606`](https://whatsonchain.com/tx/779ea67979d38deeefffa8e734125e701572582996fc51fa7bb66646f6073606) at block 949222). Without the patch, identical transactions are rejected by every ARC broadcaster with NULLFAIL.

PR with fix is opened alongside this issue.

E-Jacko:
Fixes #42.

## Summary

Two-part fix to `packages/runar-rs/src/sdk/contract.rs`:

1. **Drop the over-narrow `is_stateful` gate on the user-sig subscript trim.** Per BIP-143 / BSV consensus, when a script contains an `OP_CODESEPARATOR` before an `OP_CHECKSIG`, the sighash is computed over the subscript AFTER that separator — independent of whether the contract has state fields. The previous condition produced wrong sighashes (NULLFAIL on mainnet) for terminal methods on `StatefulSmartContract` subclasses.

2. **Replace `get_code_sep_index`'s template-adjusted offset with a script-walking helper `find_codesep_offsets`** when `code_script` is set. `from_txid` populates `constructor_args` with dummy `SdkValue::Int(0)` placeholders (the real arg bytes are already baked into the on-chain locking script), so `adjust_code_sep_offset()` computes a shift of zero and returns the wrong offset whenever the `OP_CODESEPARATOR` sits after constructor slots that expand at deploy time (e.g. `PubKey` args 1 byte → 34 bytes). The new helper walks the actual script byte-by-byte for the true on-chain offsets.

## Why this is correct

The BSV consensus rule is that any `OP_CODESEPARATOR` in the script before the `OP_CHECKSIG` resets `scriptCode` — regardless of whether the script involves state-mutation tooling. The previous condition was over-narrow.

State-mutating methods (`bid()`-style) produce byte-identical sighashes (their `is_stateful` branch was already true → the trim already ran). Only terminal methods (`close()`-style) get the new — and now-correct — behavior. For stateless `SmartContract` subclasses with no `OP_CODESEPARATOR`, `code_sep_idx < 0` so the trim doesn't run — identical to pre-patch behavior.

The script-walking helper correctly handles all BSV push opcodes (`0x01..0x4b`, `OP_PUSHDATA1`, `OP_PUSHDATA2`, `OP_PUSHDATA4`). Falls back to the legacy template-adjusted offset for synthetic / unit-test paths that have no `code_script` available.

## Changes

- `packages/runar-rs/src/sdk/contract.rs`:
  - Subscript-trim conditional (drop `is_stateful` term — the BIP-143-correct condition is `code_sep_idx >= 0` alone)
  - `get_code_sep_index` now prefers the script-walking helper when `code_script` is set
  - New `find_codesep_offsets` helper in the "Encoding helpers" section

No API surface change.

## Backwards compatibility

- State-mutating methods produce byte-identical sighashes
- Stateless `SmartContract` subclasses with no `OP_CODESEPARATOR` see no behavior change
- All 307 lib tests in `packages/runar-rs/` pass unchanged

## Mainnet validation

A test EnglishAuction was deployed, bid, and closed on BSV mainnet using the patched SDK:

- **Deploy**: [`2d4532e5…2815e6`](https://whatsonchain.com/tx/2d4532e575622511d2d0eaca411aab1ae96748947ec1d495ae07109aac2815e6) block 949121
- **Bid** (state-continuing — regression evidence; sighash unchanged): [`09814d16…6244c`](https://whatsonchain.com/tx/09814d1666d7ff14aaf6f1110c7654b6a4a5c6f287a25680ba8dd9a7c266244c) block 949122
- **Close** (terminal — primary evidence): [`779ea679…3606`](https://whatsonchain.com/tx/779ea67979d38deeefffa8e734125e701572582996fc51fa7bb66646f6073606) block 949222

Without this fix, the close transaction is rejected by every ARC broadcaster with `arc error 461 — Signature must be zero for failed CHECK(MULTI)SIG operation` (NULLFAIL). With this fix, the same transaction mines.

## Test plan

- [x] `cargo test --release --lib` in `packages/runar-rs/` — 307/307 passes
- [x] `cargo check` clean
- [x] Mainnet broadcast + WoC verification (above)

## Depends on / related

This fix targets the same root issue as #40 / #41 (deadline-based terminal methods). Without the locktime override from #41, the on-chain `extractLocktime >= deadline` assertion fails before CHECKSIG ever runs — so the locktime fix is a precondition for exercising this code path on a non-zero-deadline contract. The two PRs are otherwise independent and can land in either order.

This is the second of three fixes that together enable auction-style terminal-method contracts to mine on mainnet:

1. #41 — `nLockTime` override
2. **This PR** — terminal-method sighash subscript trim
3. Companion issue + draft PR — compiler stack-cleanup gate (cross-compiler scope, follows)

### DIFF
--- a/packages/runar-rs/src/sdk/contract.rs
+++ b/packages/runar-rs/src/sdk/contract.rs
@@ -314,10 +314,16 @@ impl RunarContract {
         let mut signatures = HashMap::new();
         let contract_utxo = prepared.contract_utxo.clone();
         for &idx in &prepared.sig_indices {
-            // Stateful: user checkSig is AFTER OP_CODESEPARATOR — trim subscript.
-            // Stateless: user checkSig is BEFORE — use full script.
+            // Per BIP-143 / BSV consensus: if the script contains an
+            // OP_CODESEPARATOR before the OP_CHECKSIG, the sighash is computed
+            // over the subscript AFTER that separator — regardless of whether
+            // the contract has state fields. Trimming was previously gated on
+            // `is_stateful`, which produced wrong sighashes for terminal methods
+            // (e.g. `Auction.close`) whose `is_stateful` flag was false on the
+            // call path despite the locking script containing an
+            // OP_CODESEPARATOR — failing with NULLFAIL on mainnet.
             let mut subscript = contract_utxo.script.clone();
-            if prepared.is_stateful && prepared.code_sep_idx >= 0 {
+            if prepared.code_sep_idx >= 0 {
                 let trim_pos = ((prepared.code_sep_idx as usize) + 1) * 2;
                 if trim_pos <= subscript.len() {
                     subscript = subscript[trim_pos..].to_string();
@@ -1416,8 +1422,37 @@ impl RunarContract {
         }
     }
 
-    /// Get the adjusted code separator index for a method.
+    /// Get the byte offset of an `OP_CODESEPARATOR` for a given method index.
+    ///
+    /// When `code_script` is set (i.e., the contract is loaded from chain, OR
+    /// the deploy script has already been built from real constructor args),
+    /// we walk the actual script and return the true on-chain byte position.
+    /// This is required because `from_txid` populates `constructor_args` with
+    /// dummy `SdkValue::Int(0)` placeholders — the real arg bytes are already
+    /// baked into the on-chain locking script — so `adjust_code_sep_offset()`
+    /// computes a shift of zero and returns the wrong offset whenever the
+    /// `OP_CODESEPARATOR` sits after constructor slots that expand at deploy
+    /// time (e.g. PubKey args = 1 → 34 bytes). The symptom of using the wrong
+    /// offset is NULLFAIL at `OP_CHECKSIG` for terminal methods.
+    ///
+    /// Falls back to the legacy template-adjusted offset for synthetic /
+    /// unit-test paths that have no `code_script` available.
     fn get_code_sep_index(&self, method_index: usize) -> i64 {
+        if let Some(ref code_script) = self.code_script {
+            if let Some(ref indices) = self.artifact.code_separator_indices {
+                let real_offsets = find_codesep_offsets(code_script);
+                if method_index < indices.len() && method_index < real_offsets.len() {
+                    return real_offsets[method_index] as i64;
+                }
+            }
+            if self.artifact.code_separator_index.is_some() {
+                let real_offsets = find_codesep_offsets(code_script);
+                if let Some(&off) = real_offsets.first() {
+                    return off as i64;
+                }
+            }
+        }
+
         if let Some(ref indices) = self.artifact.code_separator_indices {
             if method_index < indices.len() {
                 return self.adjust_code_sep_offset(indices[method_index]) as i64;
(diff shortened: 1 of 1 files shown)


icellan:
Superseded by #47, which extends this fix to all 7 tiers and merged to main (Closes #42). Thank you @E-Jacko — your commit is credited via Co-Authored-By in the merged squash. Closing this in favor of the cross-tier version.

PR #113, opened 2026-07-04
fix: five correctness fixes across runar-compiler and runar-testing

E-Jacko:
Five independent correctness fixes, each reproduced with a failing test first (the test files ship in the same commits). Four close "silently wrong" failure modes — places where the toolchain accepted input and produced wrong or unusable output with zero diagnostics; one closes a testing gap that made compiled stateful contracts unreplayable off-chain. All were hit in the course of building and mainnet-proving a cross-contract covenant on rc.1.

Each commit is atomic and cherry-pickable if you'd prefer these split into separate PRs.

## 1. `aa6517d` — compiler: validate `constructorArgs` shape and keys

`compile()`/`compileFromANF` silently baked nothing when `constructorArgs` was a positional array (the natural guess — it's what `RunarContract` and `TestSmartContract` take) or carried mistyped keys, emitting OP_0-placeholder scripts that fail opaquely at runtime. Now a diagnostic error when: (a) `constructorArgs` is an array instead of a named record; (b) a key matches no contract property; (c) a readonly property referenced by a method body remains unbaked after applying the args (checked on a DCE'd probe so unreferenced readonlys stay allowed). Placeholder compilation without `constructorArgs` is unchanged.

- Repro: `compile(src, { constructorArgs: [42n] })` on a contract with one readonly bigint — previously compiled "successfully" to a placeholder script.
- Tests: `constructor-args-validation.test.ts` (new).

## 2. `6e1a935` — testing: `outpoint`/`prevouts` overrides in `buildStatefulPreimage`

`buildStatefulPreimage` hardcoded a 36-zero-byte outpoint and `hashPrevouts = hash256(dummyOutpoint)`, so contracts that verify companion inputs via `extractHashPrevouts`/`extractOutpoint` (token-ft `merge`, any cross-input covenant) could not be exercised against the compiled Script locally. Adds optional `outpoint?: string` and `prevouts?: string[]` params. Defaults are byte-identical to previous behaviour.

- Tests: `mock-preimage-prevouts.test.ts` (new).

## 3. `4ca9c92` — testing: bake `constructorArgs` into `TestSmartContract` locking script

`TestSmartContract.fromArtifact(artifact, constructorArgs)` stored the args but executed the raw artifact script, so contracts whose readonly props are baked at deploy ran with OP_0 placeholders and failed with `OP_EQUALVERIFY failed` and no hint why. Now applies the artifact's `constructorSlots` substitution, mirroring `runar-sdk` `RunarContract.buildCodeScript` (descending byte-offset order); validates arg count against the ABI constructor when slots exist.

- Repro: any contract with `readonly x: Sha256` asserted in a method; `TestSmartContract.fromArtifact(artifact, [<hash>]).call(...)` → `OP_EQUALVERIFY failed` at PC≈3.
- Tests: `test-smart-contract-baking.test.ts` (new).

## 4. `01b82f8` — testing: implement OP_CODESEPARATOR (0xab) in ScriptVM

ScriptVM threw `Unknown or disabled opcode: 0xab`, so compiled stateful contracts (which all use codeSeparator-trimmed scriptCode) could not be replayed in the in-house VM at all — only @bsv/sdk `Spend` worked. Implements post-Genesis semantics (track offset, continue), exposes `lastCodeSeparator` (−1 when none executed; separators in non-executing branches ignored), handled in both the run loop and the step-mode path.

- Tests: `vm-codeseparator.test.ts` (new).

## 5. `ef055c0` — compiler: keep outer-scope refs alive across unrolled loops

Two related stack-lowering defects around unrolled `for`-loops:

- (a) a `const` defined before a loop and referenced inside it (including only inside a nested if-branch) was consumed by the first iteration → compile failure `Value 'X' not found on stack`;
- (b) worse: a **method param referenced after a loop** whose body also references it was consumed by the final iteration and silently lowered to an empty push (OP_0) — compilation succeeded, the env-based interpreter passed, but the emitted Script failed at runtime (`OP_SPLIT: position 153 out of range [0, 0]`). Silent interpreter↔Script divergence.

`lowerLoop` now collects outer-scope refs deeply (nested if-branches), excludes names the body itself rebinds, and protects outer refs in the final iteration whenever the enclosing scope still references them after the loop. The old silent OP_0 fallbacks in `lowerLoadParam` and `lowerLoadConst` are now **hard errors**, so any remaining consumption bug surfaces at compile time instead of as wrong on-chain bytes.

- Tests: `loop-outer-refs.test.ts` (compiler) + `loop-walk-vm.test.ts` (runar-testing; verifies the multi-input tx-walk pattern executes correctly in the Script VM for 1/2/3-input transactions).

## Test suites run

- Full TS compiler suite: **3449 tests** pass, including cross-compiler golden parity — the loop fix changes no existing goldens.
- Full `runar-testing` suite: pass.
- Full repo TS suite (`npx vitest run`, packages + conformance + examples) re-run green on this branch immediately before filing: **7,293 passed, 0 failed** (202 skipped).

## Cross-tier note (disclosure)

Fix 5 touches TypeScript stack-lowering only. The Go/Rust tiers currently **reject** the affected pattern loudly at compile time (`value 'X' not found on stack`) rather than emitting wrong bytes, so they have symptom (a) but not the dangerous symptom (b). Porting the fix to the other six tiers is offered as a follow-up; a conformance fixture exercising the post-loop-param-read pattern would then pin all seven tiers. Happy to do either in this PR if preferred.

Related open finding, deliberately **not** fixed here: loop lowering also discards non-zero start values and countdown direction (silent miscompile, filed separately as an issue) — same family, larger design decision (ANF loop node carries only `count`).

### DIFF
--- a/packages/runar-compiler/src/__tests__/constructor-args-validation.test.ts
+++ b/packages/runar-compiler/src/__tests__/constructor-args-validation.test.ts
@@ -1,0 +1,123 @@
+/**
+ * constructorArgs shape validation — `compile(opts.constructorArgs)` must
+ * reject inputs that would silently bake nothing and emit placeholder
+ * scripts that fail opaquely at runtime:
+ *
+ *  (a) positional arrays (natural guess, but keys match no property names)
+ *  (b) keys that don't match any contract property (typos)
+ *  (c) referenced readonly properties left unbaked after applying the args
+ */
+
+import { describe, it, expect } from 'vitest';
+import { compile } from '../index.js';
+
+// ---------------------------------------------------------------------------
+// Contract sources
+// ---------------------------------------------------------------------------
+
+const HASH_LOCK_SOURCE = `
+class HashLock extends SmartContract {
+  readonly hashValue: Sha256;
+
+  constructor(hashValue: Sha256) {
+    super(hashValue);
+    this.hashValue = hashValue;
+  }
+
+  public unlock(preimage: ByteString) {
+    assert(sha256(preimage) === this.hashValue);
+  }
+}
+`;
+
+const TWO_PROP_SOURCE = `
+class TwoProp extends SmartContract {
+  readonly target: bigint;
+  readonly unused: bigint;
+
+  constructor(target: bigint, unused: bigint) {
+    super(target, unused);
+    this.target = target;
+    this.unused = unused;
+  }
+
+  public check(x: bigint) {
+    assert(x === this.target);
+  }
+}
+`;
+
+const HASH = 'aa'.repeat(32);
+
+function errorMessages(result: ReturnType<typeof compile>): string {
+  return result.diagnostics
+    .filter((d) => d.severity === 'error')
+    .map((d) => d.message)
+    .join('\n');
+}
+
+// ---------------------------------------------------------------------------
+// Tests
+// ---------------------------------------------------------------------------
+
+describe('constructorArgs validation', () => {
+  it('rejects a positional array with a diagnostic error', () => {
+    const result = compile(HASH_LOCK_SOURCE, {
+      // Positional array — the shape RunarContract/TestSmartContract take,
+      // but NOT what compile() takes. Previously baked nothing silently.
+      constructorArgs: [HASH] as unknown as Record<string, string>,
+    });
+
+    expect(result.success).toBe(false);
+    expect(errorMessages(result)).toMatch(/positional array/i);
+  });
+
+  it('rejects keys that match no contract property', () => {
+    const result = compile(HASH_LOCK_SOURCE, {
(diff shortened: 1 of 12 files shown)


icellan:
Incorporated into `main`: the loop-outer-refs miscompile fix (7ed5877d), constructorArgs validation (83d76294), plus the runar-testing fixes (9498e3b7 bake constructorArgs, bb6b01a9 prevouts overrides, ae6c4e3d OP_CODESEPARATOR in ScriptVM). The compiler fixes were ported across all 7 tiers in 797c61ef — the loop-outer-refs case was a latent OP_0 miscompile in every tier, good catch. Closing.

PR #114, opened 2026-07-04
feat(sdk): typed low-level surface for cross-artifact tx assembly

E-Jacko:
**Stacked on #113** — this branch contains that PR's five fix commits plus one new commit (`abc8908`); only the last commit is new here (its tests use the `buildStatefulPreimage` prevouts support from #113). Will rebase once #113 merges so the diff collapses to the single commit.

## Problem

`call()` only composes contract inputs that share the primary artifact (the token-ft `merge` shape). Cross-contract composition — one transaction where a settlement covenant (input 0) spends alongside and verifies a token covenant (input 1) from a *different* artifact — currently requires hand-rolled assembly against private internals: the four assembly helpers are callable JS but typed private/absent from the `.d.ts`, forcing `as any` casts for legitimate use.

## Change

1. **Expose and document the existing low-level surface** (public in JS already, now typed): `getCodePartHex`, `computeOpPushTxWithCodeSep`, `getSubscriptForSigning`, `buildStatefulPrefix`, and the unlocking-script encoders `encodeArg`, `encodePushData`, `encodeScriptNumber`.
2. **New `assembleMultiContractCall(inputs, outputs, opts)`** (`src/multi-contract.ts`): builds and signs one transaction whose inputs are N covenant UTXOs from N different artifacts — per-input OP_PUSH_TX signatures, codesep-trimmed user sigs, ABI-driven unlock layout (`[_codePart?] opSig args [_changePKH _changeAmount] [_newAmount] preimage [selector]`), single-pass build (BIP-143 never covers unlocking scripts), fresh serialization on every build (avoids the @bsv/sdk `toHex()` cache trap that `0c819c9` addresses), and an **offline `Spend` dry-run gate on by default** — the assembled tx validates every input against @bsv/sdk's interpreter before it is returned, so a malformed assembly never reaches a broadcaster.

No wire-protocol primitives are touched — this is per-tier ergonomic surface, which CONTRIBUTING.md exempts from seven-SDK parity ("sync the wire bytes, not the API shape"). `canonicalJson` / envelope code paths are untouched.

## Tests

- 11 unit tests (`multi-contract-call.test.ts`) — Spend-validated offline, no node required, using two inline neutral contracts.
- 2 regtest integration tests (`multi-contract-call.regtest.test.ts`) — node **accept** (valid cross-artifact spend confirms) and node **reject** (tampered companion refused).
- Full runar-sdk suite after the change: **438 passed** (including the 13 new).

## Motivation / field evidence

This surface was extracted from a working system, not designed speculatively: the same assembly path (via this API) built a mainnet transaction in which a covenant input verified the attributes of a companion covenant input from a different artifact by parsing its parent transaction in Script — txid `8d96b792e6b6f376f0cc93728f62509e5594fd8cbab7f52d7b4ce0f52df8e0f1` (confirmed, 6,777 bytes, funding/deploys/consume all on public mainnet). The companion-verifier example PR (stacked next) documents the full pattern.

### DIFF
--- a/packages/runar-compiler/src/__tests__/constructor-args-validation.test.ts
+++ b/packages/runar-compiler/src/__tests__/constructor-args-validation.test.ts
@@ -1,0 +1,123 @@
+/**
+ * constructorArgs shape validation — `compile(opts.constructorArgs)` must
+ * reject inputs that would silently bake nothing and emit placeholder
+ * scripts that fail opaquely at runtime:
+ *
+ *  (a) positional arrays (natural guess, but keys match no property names)
+ *  (b) keys that don't match any contract property (typos)
+ *  (c) referenced readonly properties left unbaked after applying the args
+ */
+
+import { describe, it, expect } from 'vitest';
+import { compile } from '../index.js';
+
+// ---------------------------------------------------------------------------
+// Contract sources
+// ---------------------------------------------------------------------------
+
+const HASH_LOCK_SOURCE = `
+class HashLock extends SmartContract {
+  readonly hashValue: Sha256;
+
+  constructor(hashValue: Sha256) {
+    super(hashValue);
+    this.hashValue = hashValue;
+  }
+
+  public unlock(preimage: ByteString) {
+    assert(sha256(preimage) === this.hashValue);
+  }
+}
+`;
+
+const TWO_PROP_SOURCE = `
+class TwoProp extends SmartContract {
+  readonly target: bigint;
+  readonly unused: bigint;
+
+  constructor(target: bigint, unused: bigint) {
+    super(target, unused);
+    this.target = target;
+    this.unused = unused;
+  }
+
+  public check(x: bigint) {
+    assert(x === this.target);
+  }
+}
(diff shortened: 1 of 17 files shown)


E-Jacko:
Note: #128 exports the same `encodeArg` from `runar-sdk` for the mock-preimage fix — whichever lands second is a trivial rebase.

icellan:
Landed on `main` as 1e393a3d (CI green). Rebased onto post-BUG-100 main — the assembly logic and the offline @bsv/sdk `Spend` dry-run gate are unchanged; I corrected the docstrings that still described a pushed `_opPushTxSig` (BUG-100 derives the OP_PUSH_TX signature on-chain from the preimage, so nothing is pushed there), rejected null non-Sig args with an actionable message instead of encoding the ASCII bytes of "null", and dropped the computed-then-discarded opSig. 11 Spend-validated unit tests + full runar-sdk suite green. Thanks — solid work, and the confirmed mainnet tx was a nice touch. Closing as merged.

PR #115, opened 2026-07-04
feat: verification descriptors + requireBaked option + companion-verifier example

E-Jacko:
**Stacked on #113 and #114** — this branch contains those PRs' six commits plus five new commits (`45ab252`, `fe8fe26`, `d9785a1`, `988a2f1`, `179faec`, each reviewable independently); only those five are new here. Will rebase as the earlier PRs merge. Together the three PRs complete one story: a contract that verifies the attributes of a *companion input* (another covenant spent in the same transaction) by parsing that companion's parent transaction in Script — no oracle, no indexer, and now no hand-derived byte offsets.

## The gap

Writing a companion-input verifier today means hand-deriving layout facts from compiled bytes: where each constructor slot sits, how wide its push encoding is (which is value-dependent — a bigint bakes as OP_N for 1..16 but a 2-byte push at 17, silently shifting every downstream offset), how state fields are laid out, and what the contract's slot-excised template hash is. Those derivations are per-contract heroics and break invisibly when a value crosses an encoding boundary.

### `45ab252` — feat(artifact): verification descriptors

Makes the layout facts machine-readable, split by what depends on deploy-time values:

- **Value-independent (compiler emits):** `ConstructorSlot` gains optional `name`, `type`, `valueEncoding` (`data|scriptnum|bool`) and, for fixed-size types, `fixedValueByteLength` + `fixedPushHeaderBytes`. `StateField` gains byte-layout annotations (`encoding`, `byteOffset`, `byteLength`, `tailOffset`) mirroring `serializeState` exactly, driven by a new shared `STATE_FIELD_WIDTHS` table in `runar-ir-schema` that the SDK state codec now also uses — the two layouts cannot drift. New artifact field `templateDigest`: a hash256-excised-slots recipe for recomputing the slot-excised template identity hash.
- **Value-dependent (SDK resolves):** new `runar-sdk` module `slot-layout.ts` — `resolveSlotLayout(artifact, args)` (concrete offsets/lengths and encoding class, with the OP_N / OP_0 / OP_1NEGATE / pushdata boundaries handled explicitly), `computeTemplateHash(artifact, args)`, `buildResolvedCodeHex(artifact, args)`, `resolveStateLayout(artifact)`.

All fields are additive and optional — existing consumers and existing artifacts are unaffected. Tests include an SDK parity suite proving `resolveSlotLayout`/`computeTemplateHash` reproduce exactly the values an independent indexOf/two-bake-diff oracle derives, including the encoding boundaries.

### `fe8fe26` — feat(compiler): `requireBaked` compile option

A readonly property that no method references is silently eliminated from the compiled script; its value then exists nowhere in the deployed UTXO and cannot be verified on-chain by any downstream contract. `CompileOptions.requireBaked: string[]` turns that silent elimination into a compile error (names must exist, be readonly, be referenced by a method body — checked on the post-bake, pre-fold ANF so template and baked modes behave identically; belt-and-braces slot check in template mode). Deliberately a compile **option**, not a `verifiable` keyword: a new modifier would be a parser change across all nine language front-ends; the option delivers the same guarantee with a far smaller surface. 7 new tests; full compiler suite green.

### `d9785a1` — fix(sdk,testing): harden descriptor resolution, complete TestSmartContract baking

Security-review hardening before exposing the above: (a) `collectSubstitutions` asserts the template script actually carries the 1-byte OP_0 placeholder at every constructor and codeSepIndex slot offset before splicing — corrupted/mismatched artifact descriptors now throw instead of silently producing a wrong layout and wrong template hash; (b) an enriched `ConstructorSlot.name` must match the ABI constructor param at its `paramIndex` — a swapped/forged name field throws; (c) `TestSmartContract.fromArtifact` baking now also substitutes `codeSepIndexSlots` exactly as `RunarContract.buildCodeScript` does, closing the remaining test-vs-deploy byte divergence for stateful artifacts with variable-length state fields. Negative tests for each, plus byte-parity tests pinning `TestSmartContract` baking to `buildResolvedCodeHex` for template-mode and compile-baked artifacts.

### `988a2f1` — docs(examples): companion-verifier example

`examples/ts/companion-verifier/` — a two-contract teaching example of the full pattern:

- `AttributedToken.runar.ts`: a token whose immutable attributes (attestor pubkey, category enum) are baked as constructor slots and inherited byte-for-byte across splits; documents the fixed-width slot encoding rules and the "referenced attributes only" load-bearing rule (which `requireBaked` now enforces).
- `CompanionVerifier.runar.ts`: the seven normative steps — prevouts binding, positional convention, parent binding, bounded structural walk, slot-excised template identity, attribute extraction + predicate, terminal output binding — with sign-safe `bin2num` discipline throughout.
- `CompanionVerifier.test.ts`: 17 cases in the token-ft/cross-covenant conventions (TestContract, ALICE/BOB/CHARLIE keys) — accept battery plus adversarial: multi-input parents, second allowlisted attestor, rogue attestor, counterfeit template planting an allowlisted key, tampered parent tx, tampered prevouts, wrong category, min-amount, retired status, wrong vout, >3-input parent, short-script marker.
- Layout constants are derived via `resolveSlotLayout`/`computeTemplateHash` — the example dogfoods the descriptor APIs instead of hand-counting offsets.
- Registered the cross-covenant way: auto-run by CI's `npx vitest run examples/ts/`, cross-referenced from `docs/cross-covenant-pattern.md` as the stronger companion-input form of the output-reference pattern.

### `179faec` — test(decompiler): align constructor-slot expectations with descriptors

The decompiler suites asserted the bare `{ paramIndex, byteOffset }` slot shape; they now expect the enriched shape, and the decompile→recompile round-trip comparisons ignore only the `name` field (recovered sources declare synthetic `propN` names — the name is source metadata, not byte-derivable; every byte-level field still round-trips).

## Cross-tier scoping (disclosure)

The example is TS-only for now: the bounded multi-input walk needs the outer-scope-refs-across-unrolled-loops stack-lowering fix (in #113), which the Go/Rust tiers do not carry yet (they fail with `value 'inCount' not found on stack`). Tracked exclusions with justifications were added using the repo's own mechanisms — `TS_ONLY_EXAMPLES` in `cross-compiler.test.ts` and `ZIG_PORT_PENDING` in `zig-parser-examples.test.ts` — to be removed when the loop-scope fix is ported to the other tiers. Descriptor **schema** fields are additive in `runar-ir-schema` (shared JSON schema updated); non-TS compilers ignore them safely and can adopt the enrichment incrementally. Happy to port both alongside the loop fix as follow-ups.

## Tests

- Compiler descriptor emission suite (enrichment, elimination, state layout, digest recipe, JSON schema validation) + SDK parity suite: 7/7 descriptor cases.
- `requireBaked`: 7 new tests.
- Hardening: corrupted placeholder byte, out-of-range offset, swapped slot names, corrupted codeSep placeholder, un-enriched slot acceptance, byte-parity suites.
- Example: 17/17, runs in CI's examples pass.
- Full repo TS suite (`npx vitest run`, packages + conformance + examples) re-run green on this branch immediately before filing: **7,378 passed, 0 failed** (202 skipped).

## Field evidence

The pattern this enables (verifier + attributed token composed in one tx, layout facts from descriptors) is proven on public BSV mainnet — consuming tx `8d96b792e6b6f376f0cc93728f62509e5594fd8cbab7f52d7b4ce0f52df8e0f1`.

### DIFF
--- a/docs/cross-covenant-pattern.md
+++ b/docs/cross-covenant-pattern.md
@@ -90,6 +90,16 @@ const stateRoot = substr(referencedOutput, stateRootOffset, 32n);
 See `examples/ts/cross-covenant/CrossCovenantRef.runar.ts` for a complete
 working example with tests.
 
+For the stronger form — verifying attributes of a *specific companion input*
+of the current transaction by parsing its parent tx (hash-bound to the
+spending tx via the BIP-143 preimage), with contract identity proven by a
+slot-excised *template* hash instead of an exact script hash — see the
+"Verified Companion Inputs" example pair in
+`examples/ts/companion-verifier/` (`AttributedToken.runar.ts` +
+`CompanionVerifier.runar.ts`). It addresses the third limitation above:
+the parent tx binds a specific UTXO, not just a script shape, and
+per-instance attribute values survive the identity check.
+
 ## Use in BSV-EVM
 
 The bridge covenant uses this pattern to read the state covenant's latest
--- a/examples/ts/companion-verifier/AttributedToken.runar.ts
+++ b/examples/ts/companion-verifier/AttributedToken.runar.ts
@@ -1,0 +1,133 @@
+import { StatefulSmartContract, assert, checkSig } from 'runar-lang';
+import type { PubKey, Sig, ByteString } from 'runar-lang';
+
+/**
+ * AttributedToken -- a token whose immutable ATTRIBUTES are baked into the
+ * code part of its locking script, so a later covenant can verify them by
+ * inspecting the token's parent transaction in Bitcoin Script.
+ *
+ * This is the token half of the "Verified Companion Inputs" example pair
+ * (see CompanionVerifier.runar.ts for the verifier half): cross-contract
+ * composition where one covenant input verifies attributes of ANOTHER input
+ * in the same spending transaction -- no oracle, no indexer, no coordinator.
+ *
+ * Design rules that make the attributes verifiable:
+ *
+ *  - **Attributes travel in the code part.** `attestor`, `category`, and
+ *    `batchId` are readonly and baked at deploy; every child produced by a
+ *    split inherits them byte-for-byte because the children reuse the same
+ *    code part.
+ *  - **An attribute only exists on-chain if the contract references it.**
+ *    Unreferenced readonly props are eliminated by the compiler. `revoke`
+ *    references `attestor` (checkSig) and `category` (enum floor assert)
+ *    precisely so both become constructorSlots a downstream verifier can
+ *    extract. This is the load-bearing rule of the whole pattern.
+ *  - **Amount is divisible; attribution is not.** `split` conserves `amount`
+ *    across children; the attribute slots are indivisible provenance.
+ *
+ * ATTRIBUTE SLOT ENCODING (fixed-layout template requirement):
+ *  - `attestor` is a 33-byte pubkey push -- a fixed-width 33-byte slot.
+ *  - `category` is constrained to 1..16 so it always bakes as a SINGLE OP_N
+ *    opcode byte (0x51..0x60). 0 would bake as OP_0 (indistinguishable from
+ *    an unbaked placeholder); 17+ bakes as a 2-byte push, shifting every
+ *    offset after the slot. Fixed-width slots keep the verifier's offsets
+ *    compile-time constants.
+ *
+ * The verifier derives those offsets mechanically from the artifact's
+ * verification descriptors (`resolveSlotLayout` / `computeTemplateHash` in
+ * runar-sdk) -- see the test file for the derivation.
+ *
+ * State model (UTXO): mutable state (owner, amount, status) rides in an
+ * OP_RETURN tail; spending with the same code part = transfer/split.
+ * Status: 1 = Active, 3 = Retired (consumed), 4 = Revoked (attestor recall).
+ */
+class AttributedToken extends StatefulSmartContract {
+  /** Current holder. Mutable -- updated on transfer/split. */
+  owner: PubKey;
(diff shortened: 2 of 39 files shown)


icellan:
Landed on `main` as a63d3a54 (full CI green, incl. native 7-tier conformance + the execution/differential oracles). Security review found no counterfeit-companion acceptance path — the off-chain `computeTemplateHash` and the on-chain re-derivation excise byte-for-byte identically and every excised slot is independently value-checked.

Rebased onto post-#113/#114 main; folded in the review fixes:
- **encodeStateValue** now handles the canonical `boolean` type (not only the `bool` alias), so encode agrees with decode + the descriptor annotation + the on-chain 1-raw-byte layout — the "single source of truth cannot drift" property is now actually true for boolean state. Added byte-level regression tests (assert 1 raw byte, not just serialize↔deserialize self-consistency).
- **classifyEncoded** derives the scriptnum-push header width from the real push opcode instead of assuming 1 byte (OP_PUSHDATA1/2/4 for >75-byte payloads).
- **requireBaked**: documented the baked-mode scope honestly — the belt-and-braces slot check covers template mode; baked mode relies on the pre-fold reference check, which a fold-to-tautology can bypass (a robust baked-mode check needs value-survival tracking through fold+emit). Template mode, the documented deploy path, is fully covered.

Left the `AttributedToken.category <= 16n` example tweak out — it re-pins the verifier's template-identity hashes and isn't a forgery vector (out-of-range category fails the template-hash + opcode checks, fail-closed). Additive/optional schema fields; emitted hex + ANF IR unchanged, so cross-tier byte-parity is unaffected. Thanks — excellent, field-proven work. Closing as merged.

PR #125, opened 2026-07-05
fix(sdk): force uint32 in findCodesepOffsets OP_PUSHDATA4 length

### LINKED ISSUE #117: findCodesepOffsets: uint32 overflow on OP_PUSHDATA4 length → negative pushLen → infinite loop
## Summary

`findCodesepOffsets` computes the OP_PUSHDATA4 length as
`pushLen = b0 | (b1 << 8) | (b2 << 16) | (b3 << 24)`. In JavaScript, when `b3 >= 0x80` the sign
bit makes `pushLen` **negative**; the cursor advance `i += 10 + pushLen * 2` then goes negative,
and the `while (i + 2 <= len)` loop never terminates.

## Repro / field evidence

Triggered by a contract with a large (~3 KB) state blob during a stateful method call — the client
hung (observed: ~9 hours) inside the offset scan. Any deployed script containing an OP_PUSHDATA4
whose 4th length byte has the high bit set reproduces it.

## Suggested fix

Force uint32 with `(b0 | (b1 << 8) | (b2 << 16) | (b3 << 24)) >>> 0` and add a bounds check before
advancing the cursor (bail out of the scan when the push length would run past the script end).
A patched version of exactly this shape has been running in production use downstream.

E-Jacko:
> **Note on the diff:** this branch was cut from the #113 branch rather than `main`, so the Files tab currently shows #113's commits too. Only `packages/runar-sdk/src/contract.ts` (the `findCodesepOffsets` hunk) and `packages/runar-sdk/src/__tests__/codesep-offsets.test.ts` are new here. Will rebase onto `main` as soon as #113 lands so the diff collapses to this fix alone; happy to rebase onto plain `main` now instead if you'd rather review it standalone (the fix does not depend on #113).

## Problem

`findCodesepOffsets` assembles the OP_PUSHDATA4 length with signed 32-bit ORs (`b0 | (b1 << 8) | (b2 << 16) | (b3 << 24)`). When the 4th length byte is `>= 0x80` the result is negative, the cursor advance `off += 10 + pushLen * 2` goes large-negative, and the `while (off + 2 <= n)` scan never terminates. Observed in the field as a client hang (~9 hours) during a stateful method call on a contract carrying a ~3 KB state blob.

## Fix

Force uint32 with `>>> 0`, and stop the scan when a declared push length would run past the script end (malformed script). A patch of exactly this shape has been running in production use downstream. Surgical: only the OP_PUSHDATA4 branch changes; PUSHDATA1/2 cannot overflow.

## Test

`codesep-offsets.test.ts` gains three cases: a well-formed OP_PUSHDATA4 containing 0xab in its payload, the high-bit length byte (`4e00000080…` — hangs forever before the fix; the failing run had to be killed externally), and a declared length running past the script end.

## Suites run

- `packages/runar-sdk`: 428 passed, 2 skipped (vitest)
- Full TS suite (all packages + examples + conformance): 7,309 passed, 202 skipped; the single reported failure is the known local-locale Ruby-driver artifact, which passes under `LC_ALL=en_US.UTF-8` (same environment note as #113's filing)

Closes #117

### DIFF
--- a/packages/runar-compiler/src/__tests__/constructor-args-validation.test.ts
+++ b/packages/runar-compiler/src/__tests__/constructor-args-validation.test.ts
@@ -1,0 +1,123 @@
+/**
+ * constructorArgs shape validation — `compile(opts.constructorArgs)` must
+ * reject inputs that would silently bake nothing and emit placeholder
+ * scripts that fail opaquely at runtime:
+ *
+ *  (a) positional arrays (natural guess, but keys match no property names)
+ *  (b) keys that don't match any contract property (typos)
+ *  (c) referenced readonly properties left unbaked after applying the args
+ */
+
+import { describe, it, expect } from 'vitest';
+import { compile } from '../index.js';
+
+// ---------------------------------------------------------------------------
+// Contract sources
+// ---------------------------------------------------------------------------
+
+const HASH_LOCK_SOURCE = `
+class HashLock extends SmartContract {
+  readonly hashValue: Sha256;
+
+  constructor(hashValue: Sha256) {
+    super(hashValue);
+    this.hashValue = hashValue;
+  }
+
+  public unlock(preimage: ByteString) {
+    assert(sha256(preimage) === this.hashValue);
+  }
+}
+`;
+
+const TWO_PROP_SOURCE = `
+class TwoProp extends SmartContract {
+  readonly target: bigint;
+  readonly unused: bigint;
+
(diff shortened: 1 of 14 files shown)


icellan:
Incorporated into `main` as fe213bd4 (uint32 length in `findCodesepOffsets` OP_PUSHDATA4). Closing.

PR #126, opened 2026-07-05
fix(compiler): error when a contract has no public methods

### LINKED ISSUE #120: compile() silently emits an empty script when a contract has no public methods
## Summary

A contract whose methods all lack the `public` modifier compiles with `success: true`,
`diagnostics: []`, and `script: ''`. An empty locking script is never what the author meant — a
contract with no public methods has no spending entry points — but nothing says so.

## Repro

Compile any contract class whose methods omit `public`. Result: success, zero diagnostics, empty
script. (Hit in practice while porting a contract — one missing modifier, no signal.)

## Suggested fix

Emit a diagnostic ("contract has no public methods — no spending entry points; add `public` to at
least one method") from validate. Same "silent footgun" family as the `constructorArgs` validation
work in #113 — natural to review together.

E-Jacko:
> **Note on the diff:** this branch was cut from the #113 branch rather than `main`, so the Files tab currently shows #113's commits too. Only `packages/runar-compiler/src/passes/02-validate.ts` and the new cases in `02-validate.test.ts` are new here. Will rebase onto `main` as soon as #113 lands; happy to rebase onto plain `main` now if you'd rather review it standalone (the fix does not depend on #113).

## Problem

A contract whose methods all lack the `public` modifier (methods default to private) validated cleanly and compiled with `success: true`, zero diagnostics, and an **empty script** — no spending entry points and no signal why. Hit in practice via one missing modifier while porting a contract.

## Fix

`02-validate` now reports an error when no method has `visibility === 'public'`:

> Contract 'X' has no public methods — no spending entry points; add 'public' to at least one method

Same "silent footgun" diagnostic family as the `constructorArgs` shape validation from #113 (as suggested in the issue).

## Test

Three new cases in `02-validate.test.ts`: all-private methods, a contract with no methods at all, and the no-error control with one public method. Written failing-first against the unpatched validator.

## Suites run

- `packages/runar-compiler`: 3,455 passed, 8 skipped (vitest) — no existing test or conformance fixture compiles a zero-public-method contract, so nothing else moves

Closes #120

### DIFF
--- a/packages/runar-compiler/src/__tests__/02-validate.test.ts
+++ b/packages/runar-compiler/src/__tests__/02-validate.test.ts
@@ -209,6 +209,51 @@ describe('Pass 2: Validate', () => {
   });
 
   // ---------------------------------------------------------------------------
+  // Contract must have at least one public method
+  // ---------------------------------------------------------------------------
+
+  describe('no public methods', () => {
+    it('reports error when no method has the public modifier', () => {
+      const source = `
+        class Locked extends SmartContract {
+          readonly pk: PubKey;
+
+          constructor(pk: PubKey) {
+            super(pk);
+            this.pk = pk;
+          }
+
+          unlock(sig: Sig) {
+            assert(checkSig(sig, this.pk));
+          }
+        }
+      `;
+      const result = validateSource(source);
+      expect(hasError(result, 'no public methods')).toBe(true);
+    });
+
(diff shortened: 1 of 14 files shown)


icellan:
Incorporated into `main` as 72ee9265 (validate: error when a contract has no public methods) and ported to all 7 compiler tiers in e3ad6fc5. Closing.

Issue #121, opened 2026-07-04
Loop lowering silently miscompiles non-zero-start and countdown loops


E-Jacko:
## Summary

`extractLoopCount` (`passes/04-anf-lower.ts:883-919`) accepts `i = <lit>; i </<=/>/>= <lit>` and
computes `count = bound − start` (or `start − bound` for countdowns), but the ANF `loop` node
stores **only `count`** — the start value and direction are discarded. Stack lowering
(`05-stack-lower.ts:2242-2245`) and the SDK's ANF interpreter (`anf-interpreter.ts:405-411`) then
iterate `i = 0..count-1` regardless. Meanwhile the AST-level `RunarInterpreter` used by
`TestContract` executes the *true* source semantics (`interpreter/interpreter.ts:286-313`).

## Repro — verified silent divergence

```ts
for (let i = 1n; i <= 3n; i++) { sum += i; }
```

passes parse, validate, and typecheck (no pass checks start/step). Compiled Script iterates
`i = 0, 1, 2`; `TestContract` iterates `1, 2, 3`. Countdown loops (`for (i = 3n; i > 0n; i--)` —
the `i--` update is accepted by the parser) mis-lower the same way. **Tests green, chain wrong** —
the interpreter masks the miscompile, which is the most dangerous shape of compiler bug.

## Suggested fix

- Minimal/safe (small): in `lowerForStatement`, **reject** loops whose start ≠ 0 or whose condition
  op is `>`/`>=` ("loop must be `for (let i = 0n; i < N; i++)`") until the loop node carries
  start/step; mirror the guard in `02-validate` so the error carries a source location.
- Complete (medium): add `start: bigint, step: 1|-1` to the ANF loop node (schema bump across all
  seven tiers) and push `BigInt(start + i*step)`.

## Related

Same failure class as #113's outer-scope-refs loop fix (params consumed across unrolled loops
→ silent OP_0): that fix closed *param consumption*; this remains open for *iteration values*.
The differential-harness proposal (filed separately) would have caught both mechanically.

icellan:
Fixed in #136 — the ANF loop node carries `start`/`step` (non-zero-start and countdown loops now compile correctly instead of silently iterating 0..count-1), and runtime (non-literal) loop bounds are rejected with a clean diagnostic instead of a silent miscompile/panic. All 7 tiers.

PR #127, opened 2026-07-05
fix(compiler): reject non-zero-start and countdown loops instead of silently miscompiling

E-Jacko:
> **Note on the diff:** this branch was cut from the #113 branch rather than `main`, so the Files tab currently shows #113's commits too. Only the `02-validate.ts` / `04-anf-lower.ts` guards and their tests are new here. Will rebase onto `main` as soon as #113 lands; happy to rebase onto plain `main` now if you'd rather review it standalone.

## Problem

The ANF `loop` node carries only an iteration **count** — the start value and step direction are discarded — so stack lowering and the ANF interpreter always iterate `i = 0..count-1`, while the AST-level `RunarInterpreter` used by `TestContract` executes the true source semantics. `for (let i = 1n; i <= 3n; i++)` (or any countdown) therefore passes every interpreter-level test and produces different values on chain — the silent-divergence failure shape described in #121.

## Fix (reject path only)

This is the minimal/safe half proposed in #121, until the loop node carries start/step:

- `02-validate`: source-located errors for a non-zero literal start (`For loop iterator must start at 0 (got 1n) — loops compile to i = 0..count-1; offset the iterator inside the body instead`) and for `>`/`>=` conditions (countdown), alongside the existing constant-bound check.
- `04-anf-lower` `extractLoopCount`: the same guards as hard errors, for callers that skip validation. The countdown check runs first (a countdown necessarily also has a non-zero start; the direction is the more precise diagnosis). The now-unreachable `boundVal - startVal` arithmetic collapses to the zero-start case.

Zero-start counting-up loops are untouched, and **no conformance goldens move** — every fixture and example loop in the repo starts at 0 (verified by sweep before writing the guards).

## Test

Failing-first: 3 new validate cases (non-zero start, countdown, zero-start control) + 3 new anf-lower cases (throw on non-zero start, throw on countdown, `i <= 3n` still lowers with count 4).

## Suites run

- `packages/runar-compiler`: 3,458 passed, 8 skipped (vitest), including cross-compiler golden parity

Addresses the silent-miscompile hazard of #121 via the reject path; the full fix (loop node carrying start/step across the seven tiers) remains tracked in #121.

### DIFF
--- a/packages/runar-compiler/src/__tests__/02-validate.test.ts
+++ b/packages/runar-compiler/src/__tests__/02-validate.test.ts
@@ -284,6 +284,76 @@ describe('Pass 2: Validate', () => {
       // Identifier is treated as potentially const, so should pass
       expect(hasError(result, "compile-time constant")).toBe(false);
     });
+
+    it('reports error for for-loop with a non-zero start value', () => {
+      const source = `
+        class C extends SmartContract {
+          readonly x: bigint;
+
+          constructor(x: bigint) {
+            super(x);
+            this.x = x;
+          }
+
+          public m() {
+            let sum: bigint = 0n;
+            for (let i: bigint = 1n; i <= 3n; i++) {
+              sum = sum + i;
+            }
+            assert(sum > 0n);
+          }
+        }
+      `;
+      const result = validateSource(source);
+      expect(hasError(result, 'must start at 0')).toBe(true);
+    });
+
+    it('reports error for a countdown for-loop', () => {
+      const source = `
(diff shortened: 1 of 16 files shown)


icellan:
Incorporated into `main` as e4e729cf (reject non-zero-start and countdown loops in validate + a hard guard in `extractLoopCount` so the shapes are rejected loudly instead of silently miscompiled) and ported to all 7 tiers in e3ad6fc5. Closing.

PR #128, opened 2026-07-05
fix(testing): use the SDK's encodeArg for mock-preimage constructor args

### LINKED ISSUE #122: mock-preimage buildCodePart encodes constructor args differently from the SDK — corrupt scripts validated by self-consistent tests
## Summary

`encodeConstructorArg` in `runar-testing/src/mock-preimage.ts:265-270` encodes:

- bigint → `encodeNum2Bin(n, 8)` — 8 raw bytes, **no push opcode**;
- boolean → `'01'` (the SDK uses `'51'`, OP_1);
- hex string → spliced **raw with no push prefix**.

The SDK's real path (`contract.ts:1418-1424` → `encodeArg` → `encodeScriptNumber`/
`encodePushData`) and the fixed `helpers.ts` baking both produce minimal script-number/pushdata
encodings.

## Impact — verified

For any artifact with unbaked `constructorSlots`, `buildStatefulPreimage` splices non-push bytes
where the OP_0 placeholder sat, producing a locking script whose "constructor args" execute as
random opcodes and whose codePart hash never matches the on-chain one the SDK builds. Tests using
it stay **internally consistent** (same wrong codePart on both sides of the comparison), so they
pass while validating the wrong template — a test bug that manufactures false confidence.

## Suggested fix

Delete `encodeConstructorArg` and import the SDK's now-public `encodeArg` (exported in the
cross-artifact assembly PR). Add one pinning test:
`buildCodePart(artifact, args) === new RunarContract(artifact, args).buildCodeScript()` for a
bigint + ByteString + boolean slot. Longer term, a single shared encoding module (natural home:
`runar-ir-schema`, which every package already depends on) would remove the five independent
push/script-number encoder copies across runar-sdk/runar-compiler/runar-testing.

## Related

Same family as the `TestSmartContract.fromArtifact` baking fix in #113 — file
and review together.

E-Jacko:
> **Note on the diff:** this branch was cut from the #113 branch rather than `main`, so the Files tab currently shows #113's commits too. Only `packages/runar-testing/src/mock-preimage.ts`, the `encodeArg` export in `packages/runar-sdk/src/index.ts`, the `runar-testing` package.json dependency, and `mock-preimage-arg-encoding.test.ts` are new here. Will rebase onto `main` as soon as #113 lands.

## Problem

`mock-preimage.ts`'s local `encodeConstructorArg` encoded bigints as 8 raw bytes with **no push opcode**, booleans as `01` instead of OP_1 (`51`), and spliced hex strings raw with no push prefix. For any artifact with unbaked `constructorSlots`, `buildCodePart` therefore produced a locking script whose "constructor args" execute as random opcodes and whose code-part hash never matches what `RunarContract` deploys — while tests built on it stayed internally consistent (same wrong bytes on both sides of every comparison) and kept passing. Details in #122.

## Fix

Export `encodeArg` from `runar-sdk` and use it in `buildCodePart`, deleting the divergent copy. `runar-testing` gains a workspace dependency on `runar-sdk` (no cycle — runar-sdk does not depend on runar-testing). Note: #114 also exports `encodeArg` for the cross-artifact assembly work; whichever lands second is a trivial rebase.

## Test

New pinning test (failing-first — the pre-fix run shows the divergence bytes, e.g. mock `7c2a000000000000009c…019a` vs SDK `7c012a9c…519a` for the same args): `buildLockingScript(artifact, args, {})` must be byte-identical to `new RunarContract(artifact, args).getLockingScript()` for bigint + ByteString + boolean slots, plus the OP_0/OP_N-range bigint edge encodings (0n, 1n, 16n, 17n).

## Suites run

- `packages/runar-testing`: 596 passed, 91 skipped (vitest)
- `packages/runar-sdk`: 431 passed, 2 skipped

Closes #122

### DIFF
--- a/packages/runar-compiler/src/__tests__/constructor-args-validation.test.ts
+++ b/packages/runar-compiler/src/__tests__/constructor-args-validation.test.ts
@@ -1,0 +1,123 @@
+/**
+ * constructorArgs shape validation — `compile(opts.constructorArgs)` must
+ * reject inputs that would silently bake nothing and emit placeholder
+ * scripts that fail opaquely at runtime:
+ *
+ *  (a) positional arrays (natural guess, but keys match no property names)
+ *  (b) keys that don't match any contract property (typos)
+ *  (c) referenced readonly properties left unbaked after applying the args
+ */
+
+import { describe, it, expect } from 'vitest';
+import { compile } from '../index.js';
+
+// ---------------------------------------------------------------------------
+// Contract sources
+// ---------------------------------------------------------------------------
+
+const HASH_LOCK_SOURCE = `
+class HashLock extends SmartContract {
+  readonly hashValue: Sha256;
+
+  constructor(hashValue: Sha256) {
+    super(hashValue);
+    this.hashValue = hashValue;
+  }
+
+  public unlock(preimage: ByteString) {
+    assert(sha256(preimage) === this.hashValue);
+  }
+}
+`;
+
+const TWO_PROP_SOURCE = `
+class TwoProp extends SmartContract {
+  readonly target: bigint;
+  readonly unused: bigint;
+
+  constructor(target: bigint, unused: bigint) {
+    super(target, unused);
+    this.target = target;
+    this.unused = unused;
+  }
+
+  public check(x: bigint) {
+    assert(x === this.target);
+  }
+}
+`;
+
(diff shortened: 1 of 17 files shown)


icellan:
Incorporated into `main` as df3a553f (mock-preimage constructor args go through the SDK's `encodeArg`). Closing.

PR #129, opened 2026-07-05
refactor(sdk): make WalletProvider subclass-extensible (private -> protected, typed accessors)

E-Jacko:
> **Note on the diff:** this branch was cut from the #113 branch rather than `main`, so the Files tab currently shows #113's commits too. Only `packages/runar-sdk/src/providers/wallet-provider.ts`, the `deployWithWallet` getter usage in `contract.ts`, and `wallet-provider-extensibility.test.ts` are new here. Will rebase onto `main` as soon as #113 lands; happy to rebase onto plain `main` now if you'd rather review it standalone (the change does not depend on #113).

## Motivation

Every `WalletProvider` field and internal method is `private`, so nobody can subclass it cleanly:

- `runar-sdk`'s **own** `deployWithWallet` reaches in with `(walletProvider as any).wallet` / `(walletProvider as any).basket` (contract.ts) — the base class's API is insufficient even for its own SDK.
- Downstream providers that only need to reroute transport (e.g. broadcast via an overlay that gates admission, or resolve parent txs from their own index) cannot override `fetchRawTx`/`broadcastTx`/`submitToOverlay`; the two we know of ended up duplicating the entire field set and reimplementing the pipeline instead of extending it.

## Change (zero behavior change)

- Fields: `private readonly` -> `protected readonly` (all readonly — no mutation surface added).
- `fetchRawTx` / `broadcastTx` / `submitToOverlay`: `private` -> `protected`, so the base pipeline (`broadcast()`, `ensureFunding()`, EF parent assembly) dispatches through subclass overrides.
- New typed getters `walletClient` / `basketName`; `deployWithWallet` now uses them instead of the `as any` casts.

## Test

New `wallet-provider-extensibility.test.ts`, written failing-first (10 type errors against the private surface before the change — TS2341/TS2415/TS2339). It pins:

- the polymorphic dispatch: the base broadcast pipeline's EF assembly uses a subclass `fetchRawTx` override (runtime assertion with a stubbed ARC fetch);
- protected configuration reachable from a subclass (the subclass doubles as a compile-time assertion);
- the `walletClient` / `basketName` getter surface.

## Suites run

- `packages/runar-sdk`: 428 passed, 2 skipped (vitest); package `tsc --noEmit` clean

### DIFF
--- a/packages/runar-compiler/src/__tests__/constructor-args-validation.test.ts
+++ b/packages/runar-compiler/src/__tests__/constructor-args-validation.test.ts
@@ -1,0 +1,123 @@
+/**
+ * constructorArgs shape validation — `compile(opts.constructorArgs)` must
+ * reject inputs that would silently bake nothing and emit placeholder
+ * scripts that fail opaquely at runtime:
+ *
+ *  (a) positional arrays (natural guess, but keys match no property names)
+ *  (b) keys that don't match any contract property (typos)
+ *  (c) referenced readonly properties left unbaked after applying the args
+ */
+
+import { describe, it, expect } from 'vitest';
+import { compile } from '../index.js';
+
+// ---------------------------------------------------------------------------
+// Contract sources
+// ---------------------------------------------------------------------------
+
+const HASH_LOCK_SOURCE = `
+class HashLock extends SmartContract {
+  readonly hashValue: Sha256;
+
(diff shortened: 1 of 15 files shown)


icellan:
Incorporated into `main` as b6647b59 (WalletProvider private -> protected + typed accessors). Closing.

PR #158, opened 2026-08-28
Sign lattice + EC reduction sinking: p256-wallet 304,463 → 179,890 B

icellan:
Stacked on #156 (targets `feat/script-size-opt`, not `main`) because reduction sinking only
pays alongside the constant pool: the cheap subtraction references the prime twice, so
without a pooled slot it is a regression.

Implements brief Phases 4–5 for the EC codegen. Opt-in via `--ec-reduction-sinking`; inert by
default, all 72 goldens still reproduce byte-for-byte, `script-size-check` 72/72 ok.

## Results

| fixture | shipping | + pool (#156) | **+ sinking** |
|---|---:|---:|---:|
| `p256-wallet` | 958,792 | 304,463 | **179,890** (−81.2 %) |
| `p384-wallet` | 1,963,300 | 463,435 | **272,678** (−86.1 %) |
| `p384-primitives` | 1,883,767 | 447,707 | **261,534** (−86.1 %) |
| `ec-primitives` / `ec-demo` | 1,332,782 | 433,880 | **258,303** (−80.6 %) |
| `p256-primitives` | 928,219 | 296,303 | **173,640** (−81.3 %) |
| `schnorr-zkp` | 875,189 | 281,675 | **165,848** (−81.1 %) |
| `ec-unit` | 479,716 | 157,129 | **93,678** (−80.5 %) |
| `convergence-proof` | 452,386 | 147,052 | **87,302** (−80.7 %) |

The ceiling for this transformation was measured first (#156, §3.7) at 179,796 bytes on
`p256-wallet`. This lands **94 bytes off it**, so the lattice recovers essentially all of the
available win rather than a fraction of it. The residual is reductions whose dividend the
analysis cannot prove non-negative — values joined across an `OP_IF`, which stay `Unknown` by
design.

## Why a lattice rather than a rewrite

The ceiling measurement *was* the blanket rewrite. It passed **256 EC oracle assertions** —
OpenSSL signatures on both curves, `ec-on-curve-canonicity`, `ec-degenerate-add`,
`ec-mul-scalars`, `p256-p384-scalars`, `p256-p384-ecdsa-verify` — and was still wrong. The two
paths need different facts:

- **multiply / add / mulconst** need `dividend ≥ 0`. Unsigned coordinate decoding already
  gives it, so ~70 % of reductions qualify with almost no analysis.
- **subtract**'s cheap `a − b + p` form needs the strictly stronger `subtrahend < p`, which
  `OP_BIN2NUM` of 32 unsigned bytes does **not** imply — a coordinate can exceed `p` by up to
  2^32 + 977.

```
ecAdd((0, 1), (2^256 − 1, 1))
  correct : …fffffffdfffff85f
  blanket : …0001000003d0        0x1000003d0 = 2^32 + 977 = 2^256 − p
```

Reachable only through the unguarded bare builtins — `verifyECDSA_*` and `onCurve` run a
canonicity guard first — which is why no plausible-input test finds it.

## Implementation

`Dom` is a three-point lattice (`Unknown < NonNegative < Reduced`) carried as
`ECTracker.dm`, a **slot-parallel array** to `nm` rather than a name-keyed map. Names are
reused (`_fmul_prod` is written by every multiply) and the same name can be resident twice, so
a map would go stale in exactly the cases that matter. Every `nm` mutation mirrors into `dm`
with the same splice; external mutation now routes through `pushTracked` / `popTracked` /
`removeSlotAt`; and `domainOf` throws if the lengths ever differ, because a silent desync
hands a transfer function a fact about the *wrong slot* — the one failure mode that produces a
smaller script that quietly computes something else.

Transfer functions:

| producer | fact |
|---|---|
| `fieldAdd` / `fieldMul` | non-negative iff both operands are |
| `fieldSqr` | non-negative **unconditionally** (a·a ≥ 0 for any a) |
| `fieldMulConst` | keeps the operand's sign for positive `c` |
| any reduction | `Reduced` |
| decoded coordinate | `NonNegative`, never `Reduced` |
| `rawBlock`, `OP_IF` join | `Unknown` |
| group-order (`mod n`) reduction | `NonNegative` only — never `Reduced`, so a value reduced mod n can never be mistaken for one reduced mod p |

Whether to use the cheap subtraction is a **cost comparison** (`cheapSubPays`) against the
pooled push cost, not a flag — it references the prime twice, so it is only a win once the
prime is a 2-byte pick.

## Testing

`ec-reduction-sinking.test.ts` is deliberately a differential sweep over the boundary rather
than a signature check — that question was answered wrongly once already. It runs every
emitter over `0, 1, 2, p−1, p, p+1, 2^256−1, G.x`, including the full 8×8 cross product for
`ecAdd` and `p256Add`, and requires the sunk script's *result* to equal the shipping one on
every combination — including ones no valid curve point could produce. The counterexample is
pinned by name. An absolute OpenSSL oracle then re-checks accept plus seven rejection cases,
and a non-vacuity test fails if any emitter did not actually shrink.

- 4,103 compiler tests pass
- 1,685 testing / CLI / conformance tests pass
- `golden-invariance` 72/72, `script-size-check` 72/72 ok

## Not in scope

The 7-tier port. Same posture as #156: the flag ships off, and landing it for real means
porting to `compilers/{go,rust,python,ruby,zig,java}`, regenerating 9 goldens, re-stamping
`conformance/script-size-baseline.json`, and adding provenance entries.

### DIFF
--- a/conformance/runner/script-metrics.ts
+++ b/conformance/runner/script-metrics.ts
@@ -72,6 +72,7 @@ export function tsSourcePath(fixture: string, testsDir: string = TESTS_DIR): str
 export const VARIANTS: Record<string, CompileOptions> = {
   current: {},
   'ec-pool': { ecConstantPool: true },
+  'ec-sink': { ecConstantPool: true, ecReductionSinking: true },
   liveness: { schedulerMode: 'liveness' },
   both: { ecConstantPool: true, schedulerMode: 'liveness' },
 };
--- a/packages/runar-cli/src/__tests__/experimental-flags.test.ts
+++ b/packages/runar-cli/src/__tests__/experimental-flags.test.ts
@@ -128,4 +128,10 @@ describe('experimental size-optimizer flags', () => {
     const pooled = await hexWith({ ecConstantPool: true }, 'pool');
     expect(pooled).toBe(bare);
   });
+
+  it('--ec-reduction-sinking is inert on a contract with no EC operations', async () => {
+    const bare = await hexWith({}, 'nosink');
+    const sunk = await hexWith({ ecConstantPool: true, ecReductionSinking: true }, 'sink');
+    expect(sunk).toBe(bare);
+  });
 });
--- a/packages/runar-cli/src/bin.ts
+++ b/packages/runar-cli/src/bin.ts
@@ -44,6 +44,7 @@ program
   .option('--asm', 'print ASM to stdout')
   .option('--disable-constant-folding', 'disable ANF constant folding pass')
   .option('--ec-constant-pool', 'EXPERIMENTAL: pool repeated EC curve constants (changes emitted bytes)')
+  .option('--ec-reduction-sinking', 'EXPERIMENTAL: drop provably-dead sign fix-ups from EC modular reductions')
   .option('--stack-scheduler <mode>', 'EXPERIMENTAL: operand scheduling — current|liveness', 'current')
   .option('--from-ir <path>', 'compile from an ANF IR JSON file (skips parse/validate/typecheck/anf-lower)')
   .option('--hex', 'print only the script hex to stdout (no artifact JSON)')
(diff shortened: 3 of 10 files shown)


icellan:

PR #160, opened 2026-08-29
Script-size optimizer, all 7 tiers: p256-wallet 958,792 → 147,113 B (−84.7%), corpus −65.1%

icellan:
**Not for merge as-is.** This is the script-size exploration from the brief, now complete across all seven compiler tiers. Every optimization is opt-in and default output is byte-identical everywhere. §"Landing this" at the bottom says what would still have to happen.

## Headline

`conformance/tests/p256-wallet` — the fixture the brief calls its "959,592 B reference implementation" — **958,792 → 147,113 bytes (−84.7 %)**.

| stage | p256-wallet | p384-wallet | corpus |
|---|---:|---:|---:|
| shipping | 958,792 | 1,963,300 | 13,526,563 |
| + EC constant pool | 304,463 (−68.2 %) | 463,435 (−76.4 %) | 6,285,154 (−53.5 %) |
| + reduction sinking | 179,890 (−81.2 %) | 272,678 (−86.1 %) | — |
| + fixed-base comb (NIST) | **147,113 (−84.7 %)** | **223,204 (−88.6 %)** | 4,906,225 (−63.7 %) |
| + fixed-base comb (secp256k1) | — | — | **4,726,567 (−65.1 %)** |

**8.8 MB off the corpus. 43 of 72 fixtures changed, none grown.**

`p256-wallet`'s constant pushes went 697,019 → 2,116 bytes. What remains is 70.0 % stack traffic and 25.4 % arithmetic.

## Three assumptions the measurements overturned

1. **The baseline is not where the brief says.** 72.7 % of `p256-wallet` was constant pushes, and 680,850 B of it was 20,025 separate 33-byte pushes of one number — the field prime, re-pushed at every modular reduction. Pooling it into a stack slot is the single largest win in the whole stack, and it is not an algebraic optimization.
2. **Eager dead-slot retirement is not worth building.** Only 657 of 387,749 `OP_PICK`/`OP_ROLL` sites sit deeper than 16, where the depth push crosses into 2 bytes. Depths 0–2 are all one byte.
3. **A generic scheduler cannot reach P-256.** The crypto emitters bypass `05-stack-lower.ts` entirely — they emit Stack IR directly through a hand-written tracker. That is 13.4 MB of the 13.5 MB corpus. This is the architectural finding of the exercise and it is why the work split into two independent prototypes.

## All seven tiers, byte-identical

Flags in every tier: `--ec-constant-pool  --ec-reduction-sinking  --ec-fixed-base-comb`

For one `ecMulGen` contract with all three on, **all seven compilers emit the same 50,157 bytes** (down from 424,567):

| tier | parity gate | assertions | CLI |
|---|---|---|---|
| TypeScript | `conformance/ec-flag-parity/parity.test.ts` | fixture re-derived in-process | reference |
| Go | `codegen/ec_flag_parity_test.go` | 120 subtests | identical |
| Rust | `tests/ec_flag_parity_tests.rs` | 24 emitters × 4 variants | identical |
| Python | `tests/test_ec_flag_parity.py` | 48 | identical |
| Ruby | `test/codegen/test_ec_flag_parity.rb` | 24 × 4 | identical |
| Java | `codegen/EcFlagParityTest` | 3 tests over 24 × 4 | identical |
| Zig | `passes/helpers/ec_flag_parity_test.zig` | 8 tests (secp256k1 + NIST) | identical |

### Why a new gate was needed

The flags default off, so the ordinary conformance suite — which compiles with defaults — cannot see them at all. Seven tiers could each ship a *different* `--ec-constant-pool` and the suite would stay green.

That is not hypothetical. The flags change which reduction form is emitted and which addition formula each ladder round uses. A tier that ports the constant pool but not the sign lattice's `Reduced` precondition produces a script that is **smaller**, passes every test it has, and is wrong on `ecAdd((0,1), (2^256−1,1))`.

`conformance/ec-flag-parity/expected.json` pins the exact script the TS reference emits for all 24 EC emitters under all 4 flag combinations, as a byte count plus a SHA-256. It is derived, never hand-edited, and re-derived in-process by its own test so it cannot go stale.

### Six defects it caught

None findable by a tier-local test:

| tier | defect | symptom |
|---|---|---|
| Go | `cDecomposePoint` never recorded that BIN2NUM of an unsigned coordinate is `NonNegative` | NIST tier emitted a **larger** script under `--ec-reduction-sinking` (94,137 vs 90,610) |
| Go | six NIST entry points never released the pooled prime | 2 bytes each |
| Go | the three `+n` pushes in `cEmitMul` routed through the pool | strictly smaller, but the reference pushes literals — the tiers would have diverged |
| Rust | backend options built with `..Default::default()` | `--ec-constant-pool` reached the frontend and vanished; the compile succeeded and emitted the unoptimized script |
| Python | the `u2·Q` ladder was called without options | verifier at 628,923 bytes where the reference emits 319,693 — the flag applied to exactly half of it |
| Python / Ruby / Java | verifier never released its two pooled constants | 4 bytes |

The Go `NonNegative` fact and the Rust `..Default::default()` are the interesting two: both produce a compiler that works, passes its suite, and silently does not do what the flag says.

## Soundness

**The sign lattice is a sign lattice, not a modular-domain lattice.** Multiply and add need only `dividend ≥ 0`, which unsigned decoding gives for free (~70 % of reductions). Subtraction's cheap form needs `subtrahend < p` — and `OP_BIN2NUM` of 32 unsigned bytes does **not** imply that. The blanket version passed 256 EC oracle assertions and was still wrong: `ecAdd((0,1),(2^256−1,1))` differs by exactly `2^256 − p = 2^32 + 977`. `Reduced` is tracked separately for that one reason.

**The comb re-derives the interval argument rather than inheriting it.** The binary ladder uses the cheap incomplete add everywhere but its last step, justified by an interval argument over `c_i mod n` — and that comment insists the argument be redone by anything changing the offset or the iteration count. A comb changes both. `comb.ts#combSafeRounds` is that argument as executable interval arithmetic: it proves 81 of 86 rounds at w=3 and falls back to the complete add-or-double form on the rest (~1.2 kB). `combParams` searches for the scalar offset rather than reusing the ladder's `+3n` — right for P-256 and secp256k1 at w=3, **wrong for P-384**, which needs `+5n`.

**secp256k1's curve record is written out, not templated.** `makeCurve` hardcodes the NIST `a = −3`; secp256k1 is `y² = x³ + 7`. A wrong `a` there does not produce an obviously broken table — it produces a table of points on a *different* curve, which that curve's on-curve check accepts. The published 2G vector is pinned for exactly that.

**`ecMul` / `p256Mul` / `p384Mul` are deliberately not combed.** Their bases arrive at run time and the interval argument does not cover an attacker-chosen point. Asserted in the fixture test.

## Straus/Shamir: measured, then rejected

Not skipped — measured. Complete addition costs +299 B/round, taking the P-256 ladder 90,610 → 167,410 (+84.8 %). Joint-and-complete lands at ≈700 B per bit-position against 690 for two independent ladders. Straus only wins with the *incomplete* formula, whose interval argument collapses once the accumulator is `c_i·G + d_i·Q` with attacker-supplied `Q`: choose `Q = k·G` and solve `c_i + d_i·k ≡ 1 (mod n)`. The comb was taken instead because its single compile-time generator keeps the argument intact.

## A miscompile the byte counts did not catch

The liveness scheduler once accepted a witness the shipping compiler rejects. Byte counts, all 72 goldens, and 4,113 unit tests were green. `conformance/witnesses/` caught it through `runDifferentialExecution` — 2 of 86 cases. Cause: restoring alt-stack spills immediately before an `if` leaves a layout `lowerIf`'s arm reconciliation was not written for. Fix: refuse to spill in any scope with control flow ahead.

My first bisect **falsely acquitted** the spilling: with commutative reordering disabled, the method-level cost guard simply preferred the baseline schedule, so no spilling happened at all. Only after confirming the variant still changed bytes did the second bisect mean anything.

## Tier-specific findings

**Rust** pushed the EC constants as pre-encoded script-number byte blobs because they exceed `i128`. `PushValue::Int` carries a `BigInt`, so the blob was never necessary — and it cost real things: a `Bytes` push is invisible to the peephole's constant folding (which is why `k + 3n` had to be hand-folded there and only there) and invisible to the sign lattice. Switching to `Int` left the shipped bytes unchanged and brought the raw op-count goldens into agreement with Go, where they had been 4 ops short.

**Zig** carried the same duplication in `nist_ec_emitters.zig` (`NistTracker`, predating the lattice) and it was deleted the same way — the NIST emitters now share `ec_emitters.ECTracker`. Its one curve-specific field became a parameter; nothing about a tracker is per-curve.

**Rust** also carried a hand-copy of `ECTracker` in `p256_p384.rs`, commented "duplicated since it's private there". Fine for 200 lines of stack bookkeeping; not fine once it carries a sign lattice whose transfer functions decide which reduction shape gets emitted — two copies is two chances to prove `Reduced` where only `NonNegative` holds. Deleted; the tiers share one tracker.

**Zig** is the one tier that cannot assert the raw hash, for two reasons — both differences in *spelling* that its own peephole normalises away before the script ships.

*`k + 3n`.* Its peephole reassociates only `i64` `push_int` chains, and a 256/384-bit constant is a `push_data` blob in its IR, so the offset stays pre-folded there while the reference emits three `+n` steps its own peephole collapses. The two curve families differ in *when* this bites: the secp256k1 reference routes those pushes through the constant pool, so Zig matches raw-for-raw as soon as `--ec-constant-pool` is on and diverges only under `off`; the NIST reference (`cEmitMul`) pushes raw literals under **every** flag combination, so the divergence there is constant — −70 B per P-256 ladder (`3×(1+33)+3` against `(1+33)+1`) and −102 B per P-384 ladder. Copying either family's conditional into the other is a live way to get this wrong.

*MINIMALDATA on one-byte blobs.* The reference writes `push [0x02]` as `OP_2`; Zig's encoder always length-prefixes. Two such pushes per site — the `0x02`/`0x03` prefix pair in `decompressPubKey`, and the parity select in `pNNNEncodeCompressed`. This one is **pre-existing and flag-independent**: it is there with every flag off, on an emitter no flag reaches. Nothing gated it before this PR.

Zig therefore gates on the raw byte *count* with both priced exactly in `allowedDelta` — which returns zero for every emitter they do not name, and derives the ladder term from `ladderCount(name, variant) × THREE_N` rather than a table of constants, so a tier that combed `u2·Q` (an attacker-supplied base, outside the interval argument) changes the count and fails rather than absorbing the difference. Plus end-to-end hex identity through the CLI; the fixture carries a `postPeephole` measurement for the distinction.

Zig's cost model also differs by construction: its `roll`/`pick` carry the depth themselves, so they cost `sizeOfScriptNumber(depth) + 1` where every other tier charges 1 and counts a separate push.

**Ruby** needed an explicit extended-Euclid `mod_inverse`: `Integer#pow` rejects a negative exponent, so there is no `x.pow(-1, m)` shortcut.

## Verification

- **Default output unchanged everywhere.** 72/72 goldens byte-for-byte, `script-size-check` 72/72 ok, and every tier pins that flags-off reproduces the shipping hash for all 24 emitters.
- Per-tier suites: Go all green · Rust all green · Python 1,208 passed / 1 skipped · Ruby 67 test files · Java 683 tests · Zig 761 tests.
- The cost model is a *checked mirror* of the emitter, not a second opinion: `estimateScriptBytes(ops) === emitMethod(ops).scriptHex.length / 2`, swept over the whole corpus in every tier.
- Semantic equivalence under the flags: `ec-comb.test.ts` (83), `ec-constant-pool-equivalence`, `ec-reduction-sinking`, `liveness-scheduler-equivalence` including the combined `all` variant, plus the real-crypto witness corpus.

## One thing to know about the history

`gofmt -w codegen/` reformats the whole directory. Running it while porting swept 43 unrelated files (babybear, bn254, koalabear, sp1_fri, …) into the Go commit as pure whitespace churn. `fix(go): revert the gofmt sweep…` restores them; the net diff for those files is zero, but the round-trip is visible in the history.

## Landing this

Nothing here should merge as-is:

- The checked-in EC goldens were stamped under flags-off and are unchanged — correct today, but nothing regenerates them for a flags-on world, and `script-size-baseline.json` would trip its 50 % shrink guard by design.
- No golden-provenance entries exist for a flags-on stamping, because nothing has been stamped.
- The liveness scheduler is reported separately: it moves 34 fixtures but −0.0 % of corpus bytes. Its value is the *structural* guarantee (lower both ways, keep the cheaper post-peephole, so it can never grow a method), not the number.

Full write-up including the answers to all eight of the brief's questions: `docs/experiments/script-size-optimizer-results.md`. Supersedes #156, #158, #159.


icellan:
Closing: 652 commits behind, 37 semantic conflicts in the EC emitters #168 rewrote, CI already red on its own tip. The PR body said \"Not for merge as-is\" and that still holds. Do not rebase or conflict-resolve.

The measured write-up is parked in #171. Port the flags onto current emitters as three PRs (constant-pool, reduction-sinking, fixed-base comb), default off, with `ec-flag-parity`. Leave the liveness scheduler out until the oracle-price `signWith: alice` failure is understood.

Replacement (docs only): https://github.com/icellan/runar/pull/171

Issue #123, opened 2026-07-04
RFC: per-method sighash declaration — everything is pinned to SIGHASH_ALL|FORKID (0x41)


E-Jacko:
## Summary

The toolchain pins every stateful method to `SIGHASH_ALL|FORKID`: the compiler injects
`extractSigHashType(preimage) === 0x41` into every stateful public method
(`04-anf-lower.ts:163-175`), the SDK hardcodes 0x41 in `oppushtx.ts:16,79`, and the mock builder
in `mock-preimage.ts:83`. The language even exports a `SigHash` constant set
(`runar-lang/src/types.ts:280-297`) that nothing in the toolchain can use.

## Why it matters

With ALL|FORKID pinned, whole covenant families are inexpressible:

- **ANYONECANPAY** — open-participation patterns (crowdfund/join: anyone adds inputs);
- **SINGLE** — pairwise input↔output covenants, the standard way to let N token inputs each bind
  only their own continuation without agreeing on the global output set;
- **NONE** — blind-escrow shapes.

Multi-party construction today requires every covenant input to precommit to the *entire* final
output set (hashOutputs under ALL), which forces a single-coordinator assembly model.

## Proposal

Per-method sighash declaration, e.g.
`static readonly sighash = { transfer: SigHash.SINGLE | SigHash.FORKID }` (or a `/** @sighash */`
directive the parsers can read). Compiler: (1) swap the injected 0x41 constant for the declared
mode; (2) — the real work — **validate field usage against the mode**: reject
`extractHashPrevouts`/other-input reads under ANYONECANPAY, reject hashOutputs-continuation under
NONE, restrict continuation binding to the same-index output under SINGLE. SDK: thread the scope
through `computeOpPushTx` (`LocalSigner.sign` already accepts a `sigHashType` — the plumbing
exists everywhere except where it matters).

Wrong-mode covenants are a classic exploit source, which is exactly why the *validated* version
belongs in the compiler rather than making authors hand-roll it. Filed as an RFC: touches the
language surface, all seven compiler tiers, and the auto-injected preimage assert (the existing
hard-assert would need to become mode-aware).

icellan:
[...]

PR #167, opened 2026-09-09
feat(compiler): @bindingVariant directive + unified C=1 Any-S binding (7 tiers)

brendan-elas:
## What

Two coupled changes to the Any-S OP_PUSH_TX preimage-binding, ported across all seven tiers.

### 1. `@bindingVariant <lowS|all>` directive
A per-method TypeScript-surface directive (mirrors the `@sighash` mechanism from #123 end to end: parse → `MethodNode` → `check_preimage` ANF node → stack-lowering → `raw_bytes` blob; fail-closed on the eight non-TS surface parsers). Absent ⇒ default `lowS`, so golden ANF/script is unchanged for any method that doesn't use it.

### 2. Unified both constructions on a single pubkey (`038ff83d…`)
Both variants derive `s ≡ z + 1` and verify against the C=1 key; they differ only in the low-S fixup:

- **`lowS`** (default): `s = lowS((z + 1) mod n)` — branchless low-S fixup, canonical `s ≤ n/2`, accepted under the LOW_S rule (nVersion = 1). **421 bytes** (was 428 on a second pubkey `02b405d7`, now retired from the default path).
- **`all`** (opt-in): `s = z + 1` as-is (no mod-n, no low-S) — **376 bytes**, ~45 B smaller. Valid only for spends with nVersion != 1, where LOW_S is not enforced; the author owns that guarantee (the compiler can't verify a spend-time property).

## Why

One pubkey instead of two simplifies the construction, and `@bindingVariant all` lets non-`nVersion=1` covenants opt into the smaller binding. Also corrects the documented `all` failure probability: a genuine spend is rejected only when the un-reduced `s = z+1` lands `≥ n`, i.e. `(2^256 − n + 1)/2^256 ≈ 2^-128` (the previous comment claimed `2^-255`). `lowS` fails only at the single degenerate `z = n-1` (s=0), `2^-256`.

The unification changes the default binding blob, so every covenant golden was regenerated from the TS reference.

## Scope of the change

- **All 7 compiler tiers** pin the new `lowS` (421 B) and `all` (376 B) blobs and thread the directive from their TS-surface parser to codegen (fail-closed elsewhere).
- **Goldens regenerated** from the TS reference: `conformance/tests` (33 covenant fixtures), `sdk-output`, `sdk-vertical`, `analyzer`, `script-size-baseline`, and the decompiler templates.
- **Lean formal verification**: `RunarVerification/Stack/Lower.lean`'s pinned binding `ByteArray` updated to the new blob; the one moved golden anchor (`state-covenant`) re-derived live (byte-exact), fingerprint bumped, provenance updated. `pipelineGolden` passes 58/74 byte-exact.

## Testing

- Cross-tier conformance gate: **74/74 byte-identical** (all seven tiers, incl. Java).
- TS multi-format: **897 / 0**; `sdk-output` byte-identical across all seven SDKs.
- Native suites: Go / Rust / Ruby / Java / Zig green; Python via the gate.
- The binding is proven end-to-end against the BSV Script interpreter for both variants (accept genuine preimage; reject decoupled/forged/zeroed); a new `binding-failure-rate.test.ts` pins the exact failure regions and the `P = (C·Gx^-1)·G` pubkey derivation, and empirical Monte-Carlo confirms the analytic failure rates.


icellan:
Closing: 579 commits behind main, 117 conflicts, cannot merge. The 22 source conflicts are 5–23 line threading patches sitting in files #168 rewrote by 300–1,200 lines (`05-stack-lower.ts` +1216/−170 vs this PR's +11/−9). Conflict-resolving that is how you ship a `checkPreimage` that compiles and binds the wrong blob.

@brendan-elas — replayed onto current main as #170 rather than merging this branch. Three things I want you to look at there:

1. **`@bindingVariant all` vs SDK nVersion=1.** Every Runar builder uses version 1. Version > 1 is the `@bsv/sdk` relaxed regime that disables LOW_S / minimal-number / clean-stack. I kept `all` as you wrote it (author owns the nVersion guarantee) and did not invent a pin you didn't design. If you want it gated or cut, say so on #170.

2. **Default blob change is a wire-format event** (428→422, pubkey `02b405d7`→`038ff83d`). Already-broadcast UTXOs are frozen; recompiles are not.

3. **What landed on top of your patch** because #168 shipped in between: nested-context inheritance of `bindingVariant` (same as R-082 for `@sighash`), `runar-ir-schema` field (this PR only updated the compiler AST copy), goldens restamped from current main not from this dump.

Please confirm the C=1 derivation and the failure-rate correction (`all` ≈ 2⁻¹²⁸, not 2⁻²⁵⁵) on #170. Interpreter tests there agree with the rates in this PR's body.

Replacement: https://github.com/icellan/runar/pull/170

